PR #7, opened 2017-11-13
Generate mocks via go:generate directives

rupertchen:
Jira: PUB-4451

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/7)

### DIFF
--- a/build/Dockerfile.gazette-build
+++ b/build/Dockerfile.gazette-build
@@ -41,22 +41,7 @@ RUN base=github.com/LiveRamp/gazette ; \
       protoc -I src/ -I src/$base/vendor/ --plugin=/go/bin/protoc-gen-gogo \
         --gogo_out=plugins=grpc:src/ $tgt ; \
     done ; \
-    \
-    for tgt in \
-      $base/consensus/Allocator \
-      $base/consensus/KeysAPI \
-      $base/consensus/Watcher \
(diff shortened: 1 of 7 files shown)


joshk0:
Assuming this permanently obviates the need for a makefile system to build Gazette, :lgtm:
If it is in addition to the makefile system, :thumbsdown:
I would ask to make sure a CI run passes but ... we don't have that yet. :)

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/7#-:-KyrK6OY0VqEdP8jzFCe:b-d9fnhv)*

rupertchen:
This does not obviate the need for a makefile system (or other build framework), but it does obviate the need for that build system to be aware of specifics such as which interfaces should be mocked. Eventually I want to see protoc also fall under here.

The change to `Dockerfile.gazette-build` is a good example of the details of some code generation being removed from the larger build process. That same `go generate` line could eventually become the only call on that `RUN`.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/7#-:-KyrKpWRi_kYrpTG2MJX:b-6wouyu)*

PR #9, opened 2017-11-28
Revert "Generate mocks via go:generate directives (#7)"

rupertchen:
This reverts commit 4539428c518ec50ca3aa86a4e49c6b997002cf45.

We would like to have more control over the build process than
go:generate is able to offer.

Jira: PUB-4488

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/9)

### DIFF
--- a/build/Dockerfile.gazette-build
+++ b/build/Dockerfile.gazette-build
@@ -41,7 +41,22 @@ RUN base=github.com/LiveRamp/gazette ; \
       protoc -I src/ -I src/$base/vendor/ --plugin=/go/bin/protoc-gen-gogo \
         --gogo_out=plugins=grpc:src/ $tgt ; \
     done ; \
-    go generate "$base/..." ;
+    \
+    for tgt in \
+      $base/consensus/Allocator \
+      $base/consensus/KeysAPI \
+      $base/consensus/Watcher \
+      $base/consumer/ConsumerServer \
+      $base/consumer/Shard \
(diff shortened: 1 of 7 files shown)


jgraettinger:
I'd actually prefer to not revert this (@joshk0, note the project doesn't use makefiles). `go generate` is simpler for devs who are doing non-docker builds.

joshk0:
how can we address the build performance problem then?

rupertchen:
Yes. I think make offers us all the control we want to have at the cost of complexity we are willing to pay.

My perspective on go:generate after this experience is that it is an "okay" option for a project that does not want to use any other build tool, but as soon as something else is available, it loses its only advantage (ease of manual execution).

It lacks the following:
- conditional execution: The `-run` option is inconvenient to use manually and is a course filter.
- parallelization: This was intentional by design to keep it simple.

I think it would be reasonable to introduce a Makefile here.

jgraettinger:
@joshk0  I think these are independent concerns, the `Dockerfile` solution here isn't used by the non-OSS project, which can use parallelized `make` targets.

@rupertchen I'm -1 on introducing `Makefiles` at this stage. KISS. I'd rather see us start version-controlling `go generate`'d targets.

rupertchen:
@jgraettinger Fair enough. Closing this PR.

@joshk0 The performance cost of go:generate here doesn't need to affect the non-OSS project. I'll adjust its build process accordingly.

PR #12, opened 2017-12-06
Add check for os.ErrNotExist in isDir func

tylercunnion:
As written, this function will fail on newer versions of pkg/sftp, because they are now normalizing the errors from SSH errors to OS ones. This change accounts for that and restores isDir to working order.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/12)


rupertchen:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/12#-:-L0AyC7UjctPFDSUQDMY:bnfp4nl)*

PR #13, opened 2017-12-09
Avoid racing when generating multiple mocks in parallel

rupertchen:
It is possible to have a race condition when generating multiple mocks
in parallel.

One mockery process may attempt to load a package while another mockery
process is writing to a file in that same package. If the former process
reads the partially written Go file, the load, and subsequently the mock
generation, will fail.

To avoid loading and writing to the same package, this change removes
the "-inpkg" flag so files are written to a "mocks" subpackage. Upon
making this change, there were package tests with import cycles that had
to be addressed. The typical cycle looked like this:

- An interface in package "foo" being mocked references another type
  also defined in "foo". The generated mock has import path "foo/mocks"
  and imports "foo".

- A subject in package "foo" has an accompanying whitebox test also in
  "foo" and using the mock. The test has import path "foo" and imports
  "foo/mocks".

Two tactics were used to break cycles.

- The whitebox test could be converted to a blackbox test use the
  package "foo_test" while its source file did not move. This keeps the
  test near the subject and breaks the cycle.

- The package may have grown too large. Splitting the package such that
  the mocked interfaces and the types it depends on are independent of
  the code that uses them breaks the cycle.

Jira: PUB-4500

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/13)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-mock_*.go
 mocks/
 *.pb.go
 vendor
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 mock_*.go
+mocks/
 *.pb.go
 vendor
--- a/build/Dockerfile.gazette-build
+++ b/build/Dockerfile.gazette-build
@@ -35,7 +35,7 @@ COPY topic /go/src/github.com/LiveRamp/gazette/topic
 RUN base=github.com/LiveRamp/gazette ; \
     set -x ; \
     for tgt in \
-      $base/consumer/service.proto \
+      $base/consumer/service/service.proto \
       $base/recoverylog/recorded_op.proto \
     ; do \
       protoc -I src/ -I src/$base/vendor/ --plugin=/go/bin/protoc-gen-gogo \
--- a/cloudstore/common_test.go
+++ b/cloudstore/common_test.go
@@ -5,15 +5,15 @@ import (
 	gc "github.com/go-check/check"
 	"github.com/stretchr/testify/mock"
 
-	"github.com/LiveRamp/gazette/consensus"
+	"github.com/LiveRamp/gazette/consensus/mocks"
 )
 
 type CommonSuite struct {
-	keys *consensus.MockKeysAPI
+	keys *mocks.KeysAPI
 }
 
 func (s *CommonSuite) SetUpSuite(c *gc.C) {
-	s.keys = new(consensus.MockKeysAPI)
+	s.keys = new(mocks.KeysAPI)
 
 	var awsEndpoint = &etcd.Response{
 		Node: &etcd.Node{Value: testAWSEndpoint},
--- a/cmd/run-consumer/main.go
+++ b/cmd/run-consumer/main.go
@@ -11,6 +11,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/LiveRamp/gazette/consumer"
+	"github.com/LiveRamp/gazette/consumer/service"
 	"github.com/LiveRamp/gazette/gazette"
 )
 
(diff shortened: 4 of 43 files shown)


rupertchen:
Notes:

- This change is meant to be reviewed with all commits combined.
- We might be able to do better with the names of the new packages. I'm open to suggestions.

---

Review status: 0 of 43 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/13#-:-L-seM1iN88oR9twoVDD:b7pmp15)*

jgraettinger:
I don't think I can get behind this. The factorings of the `consensus` and `consumer` packages feel like we're sacrificing reader clarity and adding layers of indirection purely in service of the Almighty Code Generator, and it's too high a price to pay.

I propose the following instead:

  * We instrument `mock_Foo.go` and also protobuf `foo.pb.go` targets as `go:generate` directives.
  * _We remove `go generate` from the build phase_, treating it instead as an explicit developer-driven activity. We don't even have a hard requirement that `go generate` be used as-versioned; conceptually the directives are living documentation that may need to be tweaked (eg, on windows it's `protoc.exe`, not `protoc`).
  * We remove these targets from `.gitignore` and begin explicitly versioning them.

Rationale:

  * We've had what feels like a constant nag of paper cuts from maintenance of code-generation in the build phase. We're spending too much time appeasing our tools.
  * Reviewing generated code, and understanding the details of how it's changing is a _Good Thing_. Just because it's generated doesn't mean it's not part of our build, and we're not responsible for when it breaks. Specifically, seeing how mocks or `pb.go`'s are changing on, say, `protoc-gen-gogo` upgrades is hard to do now and would be improved by this flow.
   * Relatedly, it simplifies release management handling of _our_ code vs _generated_ code. It's easy to spot what's changing, and to revert.
   * Generated code actually changes quite rarely. Manual runs don't seem a big deal.
   * It improves developer ergonomics & UX / brings us closer to being `go get`-able. I'm trying to develop on windows right now, and would love to not have to set up a protobuf compiler / mockery in order to get a build going.
   * Version-controlling generated code is common in the broader ecosystem (eg, Etcd / K8). We're unusual for doing it in this way (my bad, I started us down this path).

---

Review status: 0 of 43 files reviewed at latest revision, 1 unresolved discussion.

---

*[cmd/run-consumer/main.go, line 45 at r1](https://reviewable.io:443/reviews/liveramp/gazette/13#-L-wjfriWt2hJ3Mm8Rx2:-L-wjfriWt2hJ3Mm8Rx3:b-1xejy4) ([raw file](https://github.com/liveramp/gazette/blob/96735e39aa615efa9f1be55ce7b02be5e63a28ff/cmd/run-consumer/main.go#L45)):*
> ```Go
> 	flag.Parse() // Parse again to initialize any plugin flags.
> 
> 	var instance service.Consumer
> ```

still reading, but this seems off to me. this is an interface, and should be of the `consumer` package.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/13#-:-L-wmCDeZcDjbEM-SC6I:b2c3kf0)*

rupertchen:
After revisiting this after the weekend, I think you're right. I agree with all three bullets of the proposal with the addition that the new mocks be generated with `-testonly` so the output files are `*_test.go` files.

I'll close out this PR and draft up a new change.

---

Review status: 0 of 43 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/13#-:-L05fI02T_RDJPdkWrkG:bhn1v0s)*

PR #20, opened 2018-01-15
Breaking changes

jgraettinger:
These are two changes which require updates to our internal project, but I believe are worth doing:

 * Re-factor code layout to adopt `pkg` and `cmd` conventions. This makes for a cleaner project, but requires re-writes of import paths.

 * Change the signature of `topic.Publish` to return AsyncAppend. This is required for consumers who wish to be vigorous about ensuring no published messages are dropped (eg, one would wait for all results to signal prior to returning from `Consumer.Flush`).

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/20)

### DIFF
--- a/examples/word-count/counter/counter.go
+++ b/examples/word-count/counter/counter.go
@@ -55,7 +55,7 @@ func (counter) Flush(s consumer.Shard, pub *topic.Publisher) error {
 		writeBatch.Put([]byte(word), []byte(strconv.AppendInt(nil, int64(count), 10)))
 
 		// Publish the updated word count to the output topic.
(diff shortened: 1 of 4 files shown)


rupertchen:
:lgtm:

I started "reviewing" files before I noticed the review mode was "combine commits". Hopefully this has no effect on how you want to merge these changes.

---

Reviewed 301 of 301 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/20#-:-L30Q6GlI4jxeC7CRk_n:b-mc5eyt)*

PR #22, opened 2018-01-15
WriteService: remove "stop-writes" disk monitor and semantics

jgraettinger:
Stop-writes attempts to monitor local disk, and if the disk
fills beyond a threshold will block all attempts to write to Gazette
(which requires local spooling) until disk usage has dropped.

This behavior has bitten us more than it's helped, because stop-writes
often results in ensemble deadlocks which can only be resolved by
letting writes proceed. It also often impacts (sometimes important)
services which are innocent bystanders and not the primary culprits
of disk usage.

Alternatives include monitoring at the application level,
implementing quotas on disks assigned to applications, and likely other
approaches.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/22)

### DIFF
--- a/pkg/gazette/write_service.go
+++ b/pkg/gazette/write_service.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"os"
 	"sync"
-	"syscall"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -21,8 +20,6 @@ var (
 	// Time to wait in between broker write errors. Exposed for debugging.
 	writeServiceCoolOffTimeout = time.Second * 5
 
-	diskUsageThreshold = flag.Int("diskUsageThreshold", 75,
-		"HWM in % at which to block further writes to Gazette.")
 	writeConcurrency = flag.Int("gazetteWriteConcurrency", 4,
 		"Concurrency of asynchronous, locally-spooled Gazette write client")
 )
@@ -91,14 +88,6 @@ type WriteService struct {
 	// Indexes pendingWrite's which are in |writeQueue|, and still append-able.
 	writeIndex   map[journal.Name]*pendingWrite
 	writeIndexMu sync.Mutex
-
-	// RWMutex used in the following way:
-	// - Calls to obtainPendingWrite lock it for READ.
-	// - A disk usage checker goroutine will lock it for WRITE if disk usage
-	//   exceeds a stated threshold, causing calls to obtainPendingWrite to
-	//   block until the condition is resolved. The goroutine logs at ERROR
-	//   until the condition is resolved, so it is easy to diagnose.
-	diskUsageMu sync.RWMutex
 }
 
 func NewWriteService(client *Client) *WriteService {
@@ -130,7 +119,6 @@ func (c *WriteService) Start() {
 		panic(err)
 	}
 
-	go c.monitorDiskSpace()
 	for i := range c.writeQueue {
 		go c.serveWrites(i)
 	}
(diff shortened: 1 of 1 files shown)


joshk0:
I'm OK with this code-wise, but feel like the burden is on you to flesh out those monitoring alternatives before I can say this is ready to go. What do you think?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/22#-:-L2vVALtNlyDZJi38R8g:bflg0cx)*

jgraettinger:
It's an open internal task, but I also don't think it falls to this project to own (since gazette is technically agnostic to how it's orchestrated & managed in a cluster)

Our experience with this implementation to-date has me convinced that no implementation is preferable, due to the unintended consequences we've observed.

joshk0:
OK, so let me go through your proposed alternatives first.

> monitoring at the application level

in other words, move the disk usage check to above the gazette abstraction level? Seems reasonable to me, though slightly repetitive. I also don't think it will change what happens when we do run out of disk space.

> implementing quotas on disks assigned to applications

This is currently not possible in Kubernetes and isn't expected to be possible very soon. You can only restrict total use of PersistentVolumes, like [this](https://kubernetes.io/docs/tasks/administer-cluster/limit-storage-consumption)

Stepping back, the initial impetus for making this change was making disk pressure issues less likely to cause a dataloss scenario and i think it succeeded in that regard. Before, it put a lot of stress on Ops to police storage usage so that the fatal condition is not hit. I cannot help but feel that reverting this change brings us back to that state of things.

I agree that the implementation of stop-writes as it stands was not great, and that a throttling approach based on disk pressure could be better. However, I feel kind of naked with the change backed out altogether. I guess I would be OK with it if gazette OC shares responsibility for disk usage concerns throughout the cluster.

jgraettinger:
https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/

This seems about what we want.

joshk0:
Unfortunately, this is complicated by our use of the local SSD (the
constrained resource):

>If you store volumes and logs in a dedicated file system, the node will
not monitor that file system.

I am not sure eviction thresholds will work, but I agree we should confirm
these concerns with experimentation.

jgraettinger:
Looks like this already all works. Poking at a `kubelet` and hitting it's `/stats/summary`, I see (among other things):

>    "fs": {
    "time": "2018-01-18T04:28:31Z",
    "availableBytes": 337717010432,
    "capacityBytes": 395260067840,
    "usedBytes": 37393620992,
    "inodesFree": 24574980,
    "inodes": 24576000,
    "inodesUsed": 1020
   },
   "runtime": {
    "imageFs": {
     "time": "2018-01-18T04:28:31Z",
     "availableBytes": 87427870720,
     "capacityBytes": 101258067968,
     "usedBytes": 15200228552,
     "inodesFree": 4811766,
     "inodes": 6258720,
     "inodesUsed": 1446954
    }
   }

Eg, the `nodefs` volume it's monitoring is the volume mounted under the `/var/lib/kubelet` path. Source diving confirms this, and usage lines up with `df` of `/dev/sdb`.

I also see that hard evictions are already enabled:

> /home/kubernetes/bin/kubelet ...  --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%

That limit is lower than the stop-writes threshold, which explains why we've never noticed it being hit.

I think we're already covered here. The pony would be to additionally enable soft evictions.

joshk0:
Fair, but how do we prevent Gazette from being evicted first? IIRC that's what happened before which caused us data loss.

joshk0:
nm, maybe we can do [this](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed)

jgraettinger:
I believe we've only seen memory-based evictions of Gazette, and that was entirely due to the size of the page cache it's container was using.

I don't agree with the premise that we _should_ be preventing Gazette from being evicted first, especially due to disk usage.

joshk0:
Discussed in meatspace. Assuming the following:

* Hard evictions are already working in Kubernetes against the local SSD
* The eviction algorithm understands usage of disk by each pod
* The eviction algorithm favors killing the largest user of disk and that's most often not Gazette but consumers

Then I am fine relying on built in eviction mechanism and this :lgtm:

PR #23, opened 2018-01-18
Use Kubernetes Service dns for service discovery.

julianvmodesto:
Prefer Kubernetes Service dns over env vars for service discovery
because env vars aren't available across Namespaces or for ExternalName
Services, and env vars are required at Pod startup, whereas dns is more
resilient to Endpoint updates.

Remove 127.0.0.1 as the default for discovering etcd and gazette
and expect Kubernetes first. Finally, remove _SERVICE_HOST and
_SERVICE_PORT since this will clobber the dns service discovery in favor
of the env vars, which we don't prefer.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/23)

### DIFF
--- a/envflag/envflag.go
+++ b/envflag/envflag.go
@@ -7,30 +7,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strings"
 )
 
 type parseableFromEnv interface {
 	parseFromEnv()
-}
-
-// serviceEndpointFlag holds the metadata for an envflag representing a service
-// endpoint. A service endpoint is a host-port pair.
-type serviceEndpointFlag struct {
-	// host and port are the pair of names of the environment variables.
-	host, port string
-	// ptr is the address of the string variable that stores the value of the
-	// flag. This typically is the return value of a flag.String call.
-	ptr *string
-}
-
-// parseFromEnv implements parseableFromEnv
-func (f serviceEndpointFlag) parseFromEnv() {
-	if h := os.Getenv(f.host); h != "" {
-		if p := os.Getenv(f.port); p != "" {
-			*f.ptr = h + ":" + p
-		}
-	}
 }
 
 // stringFlag holds the metadata for an envflag representing a simple string. A
@@ -61,21 +41,6 @@ type FlagSet struct {
 // NewFlagSet returns a new, empty flag set.
 func NewFlagSet(fs *flag.FlagSet) *FlagSet {
 	return &FlagSet{fs: fs}
-}
-
-// ServiceEndpoint defines a service endpoint flag. A service endpoint is a
-// host-port value, parsed either from a pair of environment variables
-// "<NAME>_HOST", "<NAME>_PORT" or a single colon-delimited
-// command-line argument "-<name>Endpoint". The return value is the address
-// of a string variable that stores the value of the flag.
-func (fs *FlagSet) ServiceEndpoint(name, value, usage string) *string {
-	var prefix = strings.ToUpper(name)
-	var evHost = prefix + "_HOST"
-	var evPort = prefix + "_PORT"
-
-	var ptr = fs.fs.String(name+"Endpoint", value, fmt.Sprintf("%s (%s, %s)", usage, evHost, evPort))
-	fs.addFlag(name, serviceEndpointFlag{evHost, evPort, ptr})
-	return ptr
 }
 
 // String defines a string flag. A string flag is a single string value, parsed
--- a/envflag/envflag.go
+++ b/envflag/envflag.go
@@ -65,13 +65,13 @@ func NewFlagSet(fs *flag.FlagSet) *FlagSet {
 
 // ServiceEndpoint defines a service endpoint flag. A service endpoint is a
 // host-port value, parsed either from a pair of environment variables
-// "<NAME>_SERVICE_HOST", "<NAME>_SERVICE_PORT" or a single colon-delimited
-// command-line argument "-oss<name>Endpoint". The return value is the address
+// "<NAME>_HOST", "<NAME>_PORT" or a single colon-delimited
+// command-line argument "-<name>Endpoint". The return value is the address
 // of a string variable that stores the value of the flag.
 func (fs *FlagSet) ServiceEndpoint(name, value, usage string) *string {
 	var prefix = strings.ToUpper(name)
-	var evHost = prefix + "_SERVICE_HOST"
-	var evPort = prefix + "_SERVICE_PORT"
+	var evHost = prefix + "_HOST"
+	var evPort = prefix + "_PORT"
 
 	var ptr = fs.fs.String(name+"Endpoint", value, fmt.Sprintf("%s (%s, %s)", usage, evHost, evPort))
 	fs.addFlag(name, serviceEndpointFlag{evHost, evPort, ptr})
@@ -80,7 +80,7 @@ func (fs *FlagSet) ServiceEndpoint(name, value, usage string) *string {
 
 // String defines a string flag. A string flag is a single string value, parsed
 // either from an environment variable "<envvarName>" or a command-line
-// argument "-oss<flagName". The return value is the address of a string
+// argument "-<flagName". The return value is the address of a string
 // variable that stores the value of the flag.
 //
 // This is a general flag creation utility which is why flagName and
--- a/envflag/envflag_test.go
+++ b/envflag/envflag_test.go
@@ -17,27 +17,6 @@ type envflagSuite struct {
 
 var _ = check.Suite(&envflagSuite{})
 
-func (s *envflagSuite) TestServiceEndpoint(c *check.C) {
-	var fs = flag.NewFlagSet("TestServiceEndpoint", flag.ContinueOnError)
-	var efs = NewFlagSet(fs)
-	var sut = efs.ServiceEndpoint("dummyName", "default value", "Foo bar baz")
-
-	// Envflags create underlying flags.
-	var actualFlag = fs.Lookup("dummyNameEndpoint")
-	c.Assert(actualFlag, check.NotNil)
-	c.Check(actualFlag.DefValue, check.Equals, "default value")
-	c.Check(actualFlag.Usage, check.Equals, "Foo bar baz (DUMMYNAME_HOST, DUMMYNAME_PORT)")
-
-	// Service endpoint flags parse from a pair of environment variables
-	defer assertAndSetenv(c, "DUMMYNAME_HOST", "dummy.example")()
-	defer assertAndSetenv(c, "DUMMYNAME_PORT", "1234")()
-
-	// Verify pre- and post-Parse values.
-	c.Check(*sut, check.Equals, "default value")
-	efs.Parse()
-	c.Check(*sut, check.Equals, "dummy.example:1234")
-}
-
 func (s *envflagSuite) TestString(c *check.C) {
 	var fs = flag.NewFlagSet("TestString", flag.ContinueOnError)
 	var efs = NewFlagSet(fs)
--- a/envflag/envflag_test.go
+++ b/envflag/envflag_test.go
@@ -26,11 +26,11 @@ func (s *envflagSuite) TestServiceEndpoint(c *check.C) {
 	var actualFlag = fs.Lookup("dummyNameEndpoint")
 	c.Assert(actualFlag, check.NotNil)
 	c.Check(actualFlag.DefValue, check.Equals, "default value")
-	c.Check(actualFlag.Usage, check.Equals, "Foo bar baz (DUMMYNAME_SERVICE_HOST, DUMMYNAME_SERVICE_PORT)")
+	c.Check(actualFlag.Usage, check.Equals, "Foo bar baz (DUMMYNAME_HOST, DUMMYNAME_PORT)")
 
 	// Service endpoint flags parse from a pair of environment variables
-	defer assertAndSetenv(c, "DUMMYNAME_SERVICE_HOST", "dummy.example")()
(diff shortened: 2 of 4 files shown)


joshk0:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[envflag/envflag.go, line 73 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L39s_cDYdI70sUruaOm:-L39s_cDYdI70sUruaOn:b-ilm6tx) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflag/envflag.go#L73)):*
> ```Go
> func (fs *FlagSet) ServiceEndpoint(name, value, usage string) *string {
> 	var prefix = strings.ToUpper(name)
> 	var evHost = prefix + "_HOST"
> ```

where are we passing that the default endpoints are e.g. etcd:2379 and redis:6379? i don't see that in this change.

I would say that if we don't want to absorb *_SERVICE_{HOST,PORT} anymore we should just scrap the functionality entirely.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23)*

julianvmodesto:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[envflag/envflag.go, line 73 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L39s_cDYdI70sUruaOm:-L39vWc1cHP-_ms510fu:b-i431c3) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflag/envflag.go#L73)):*
> where are we passing that the default endpoints are e.g. etcd:2379 and redis:6379? i don't see that in this change.

Done in envflagfactory/factory.go below.

> I would say that if we don't want to absorb *SERVICE{HOST,PORT} anymore we should just scrap the functionality entirely.

I kept the env var config for flexibility, so users can flag parse or env var parse or both in the order they want.

For example, the config below in this package is configurable by flags and/or env vars:

```
// also in envflagfactory/factory.go

// NewCloudFSURL defines the cloudFS URL flag.
func NewCloudFSURL() *string {
	return envflag.CommandLine.String(
		"cloudFS",
		"CLOUD_FS_URL",
		"file:///cloud-fs",
		"URL parameterizing the cloud filesystem to use.")
}

// NewLogLevel defines the logrus log-level flag.
func NewLogLevel() *string {
	return envflag.CommandLine.String(
		"logLevel",
		"LOG_LEVEL",
		"info",
		"Logrus log-level.",
	)
}

// NewMetricsPath defines the HTTP resource path on which to fetch metrics.
func NewMetricsPath() *string {
	return envflag.CommandLine.String(
		"metricsPath",
		"METRICS_PATH",
		"/metrics",
		"The HTTP resource path on which to fetch metrics.",
	)
}

// NewMetricsPort defines the metrics port flag.
func NewMetricsPort() *string {
	return envflag.CommandLine.String(
		"metricsPort",
		"METRICS_PORT",
		":8090",
		"Metrics port.",
	)
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23)*

joshk0:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[envflag/envflag.go, line 73 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L39s_cDYdI70sUruaOm:-L3ANrzhVjnB0Id0zbOj:bclbke0) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflag/envflag.go#L73)):*
<details><summary><i>Previously, julianvmodesto (Julian V. Modesto) wrote…</i></summary><blockquote>

> where are we passing that the default endpoints are e.g. etcd:2379 and redis:6379? i don't see that in this change.

Done in envflagfactory/factory.go below.

> I would say that if we don't want to absorb *SERVICE{HOST,PORT} anymore we should just scrap the functionality entirely.

I kept the env var config for flexibility, so users can flag parse or env var parse or both in the order they want.

For example, the config below in this package is configurable by flags and/or env vars:

```
// also in envflagfactory/factory.go

// NewCloudFSURL defines the cloudFS URL flag.
func NewCloudFSURL() *string {
	return envflag.CommandLine.String(
		"cloudFS",
		"CLOUD_FS_URL",
		"file:///cloud-fs",
		"URL parameterizing the cloud filesystem to use.")
}

// NewLogLevel defines the logrus log-level flag.
func NewLogLevel() *string {
	return envflag.CommandLine.String(
		"logLevel",
		"LOG_LEVEL",
		"info",
		"Logrus log-level.",
	)
}

// NewMetricsPath defines the HTTP resource path on which to fetch metrics.
func NewMetricsPath() *string {
	return envflag.CommandLine.String(
		"metricsPath",
		"METRICS_PATH",
		"/metrics",
		"The HTTP resource path on which to fetch metrics.",
	)
}

// NewMetricsPort defines the metrics port flag.
func NewMetricsPort() *string {
	return envflag.CommandLine.String(
		"metricsPort",
		"METRICS_PORT",
		":8090",
		"Metrics port.",
	)
}
```
</blockquote></details>

I still feel like it's safe to get rid of but will defer to @rupertchen on it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23)*

rupertchen:
I think this is a good opportunity to deprecate or remove `envflag.FlagSet.ServiceEndpoint()`. Details in the comments below.

---

Reviewed 1 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[envflag/envflag.go, line 73 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L39s_cDYdI70sUruaOm:-L3AkC0ava_DyZ7DUceV:b-64l12o) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflag/envflag.go#L73)):*
<details><summary><i>Previously, joshk0 (Joshua Kwan) wrote…</i></summary><blockquote>

I still feel like it's safe to get rid of but will defer to @rupertchen on it
</blockquote></details>

Long term, we should support [the twelve-factor methodology](https://12factor.net/). Given that, I think we should continue to support env vars in some form.

Given:
- the underlying goal here is to not use K8s's service env var pairs
- the original purpose of `envflag.FlagSet.ServiceEndpoint()` was to support K8s's service env var pairs

I think we should replace usages of `ServiceEndpoint()` with `String()` and use a single endpoint config variable (as opposed to the host-port pair). This is the approach taken in [upcoming changes](https://github.com/LiveRamp/gazette/blob/stream-sum/cmd/run-consumer/main.go#L34).

AFAIK, `envflagfactory.NewEtcdServiceEndpoint()` and `envflagfactory.NewGazetteServiceEndpoint()` are the only usages of `envflag.serviceEndpointFlag`, so we might be able to remove that flag type entirely.

The goal of `envflag` was to support the extraction of the gazette codebase without breaking compatibility. I expect us to deprecate the package in favor of using [Cobra](https://github.com/spf13/cobra) and [Viper](https://github.com/spf13/viper).

---

*[envflagfactory/factory.go, line 10 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L3ApDXSmAIdFNFtVofN:-L3ApDXSmAIdFNFtVofO:bplpclh) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflagfactory/factory.go#L10)):*
> ```Go
> 
> // NewGazetteServiceEndpoint defines the gazette service endpoint flag.
> // By default, use Kubernetes DNS for service discovery.
> ```

I don't think this should give Kubernetes DNS as the reason for changing the defaults. That just happens to be something we can leverage. It's more important that these defaults either are what would commonly be used or provide good documentation as an example.

I don't feel too strongly about this, but think there's an argument for the default values being purely for example and effectively requiring everyone to supply their own is the safest option and a reasonable demand. To that end, I suggest using a [reserved TLD](https://tools.ietf.org/html/rfc2606#section-2) and defaulting to `gazette.example:8081`.

@jgraettinger, your thoughts here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23#-:-L3AtMyeQQIUhsqlg8xc:bute3sh)*

joshk0:
Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23)*

julianvmodesto:
PTAL @joshk0 @rupertchen

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[envflag/envflag.go, line 73 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L39s_cDYdI70sUruaOm:-L3DpE_LUSvV1G-vTmqh:b-g1hs46) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflag/envflag.go#L73)):*
<details><summary><i>Previously, rupertchen (Rupert Chen) wrote…</i></summary><blockquote>

Long term, we should support [the twelve-factor methodology](https://12factor.net/). Given that, I think we should continue to support env vars in some form.

Given:
- the underlying goal here is to not use K8s's service env var pairs
- the original purpose of `envflag.FlagSet.ServiceEndpoint()` was to support K8s's service env var pairs

I think we should replace usages of `ServiceEndpoint()` with `String()` and use a single endpoint config variable (as opposed to the host-port pair). This is the approach taken in [upcoming changes](https://github.com/LiveRamp/gazette/blob/stream-sum/cmd/run-consumer/main.go#L34).

AFAIK, `envflagfactory.NewEtcdServiceEndpoint()` and `envflagfactory.NewGazetteServiceEndpoint()` are the only usages of `envflag.serviceEndpointFlag`, so we might be able to remove that flag type entirely.

The goal of `envflag` was to support the extraction of the gazette codebase without breaking compatibility. I expect us to deprecate the package in favor of using [Cobra](https://github.com/spf13/cobra) and [Viper](https://github.com/spf13/viper).
</blockquote></details>

Makes sense, done.

---

*[envflagfactory/factory.go, line 10 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L3ApDXSmAIdFNFtVofN:-L3DpkczXi_7oiQ5GICF:bu3g2tw) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflagfactory/factory.go#L10)):*
<details><summary><i>Previously, rupertchen (Rupert Chen) wrote…</i></summary><blockquote>

I don't think this should give Kubernetes DNS as the reason for changing the defaults. That just happens to be something we can leverage. It's more important that these defaults either are what would commonly be used or provide good documentation as an example.

I don't feel too strongly about this, but think there's an argument for the default values being purely for example and effectively requiring everyone to supply their own is the safest option and a reasonable demand. To that end, I suggest using a [reserved TLD](https://tools.ietf.org/html/rfc2606#section-2) and defaulting to `gazette.example:8081`.

@jgraettinger, your thoughts here?
</blockquote></details>

Hmm... since we'd like to have `etcd` and `gazette` be the default, I'll move the defaulting to `etcd` and `gazette` back to our internal endpoints package.

I've made this change in my downstream branch and set `gazette.example` and `etcd.example` here as suggested.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23#-:-L3Dp4unikEf7XNDYrT5:b4zw2d7)*

joshk0:
:lgtm: from my side, but i was just an interloper here. my concern from earlier was the special casing of *_HOST and *_PORT using a special set of methods.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23#-:-L3DyVtSWcdEuHKfm8Gg:bcce3ia)*

rupertchen:
:lgtm:

Thanks for taking care of a bunch of outdated comments as well!

---

Reviewed 4 of 4 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23#-:-L3E7E9gzEIMn5thuAC1:bu7u7my)*

jgraettinger:
:lgtm: also!

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[envflagfactory/factory.go, line 10 at r1](https://reviewable.io:443/reviews/liveramp/gazette/23#-L3ApDXSmAIdFNFtVofN:-L3EBzu1CK7SOJ_25hhd:b-ixf3ij) ([raw file](https://github.com/liveramp/gazette/blob/d068fbded19a732c33a516f700fe52ba0b06ecd5/envflagfactory/factory.go#L10)):*
<details><summary><i>Previously, julianvmodesto (Julian V. Modesto) wrote…</i></summary><blockquote>

Hmm... since we'd like to have `etcd` and `gazette` be the default, I'll move the defaulting to `etcd` and `gazette` back to our internal endpoints package.

I've made this change in my downstream branch and set `gazette.example` and `etcd.example` here as suggested.
</blockquote></details>

+1 from me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/23#-:-L3EC3B0Og9FeVid0X7Z:b-861wid)*

PR #40, opened 2018-01-29
Use a periodic full refresh in RetryWatcher.

julianvmodesto:
Addresses https://github.com/LiveRamp/gazette/issues/24.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/40)

### DIFF
--- a/pkg/consensus/retry_watcher.go
+++ b/pkg/consensus/retry_watcher.go
@@ -1,26 +1,29 @@
 package consensus
 
 import (
+	"time"
+
 	etcd "github.com/coreos/etcd/client"
 	"golang.org/x/net/context"
 )
 
-// RetryWatcher() composes Get() and Watch() of etcd.KeysAPI to provide a
+// RetryWatcher composes Get() and Watch() of etcd.KeysAPI to provide a
 // etcd.Watcher implementation with builtin retry for watch errors.
 // RetryWatcher differs from KeysAPI.Watcher() in two key ways:
 //  * WatcherOptions.AfterIndex is ignored. Instead, RetryWatcher() performs
 //    its own Get() on the first call to Next().
 //  * Watch-related errors will be silently retried via a Get(), which is both
 //    passed through and also used to re-establish watch consistency.
-//    Callers must be able to handle an occasional "get" response.
+//    Callers must be able to handle a periodic "get" response.
 func RetryWatcher(keysAPI etcd.KeysAPI, key string, getOpts *etcd.GetOptions,
-	watcherOpts *etcd.WatcherOptions) etcd.Watcher {
+	watcherOpts *etcd.WatcherOptions, refreshTicker <-chan time.Time) etcd.Watcher {
 
 	return &retryWatcher{
-		keysAPI:   keysAPI,
-		key:       key,
-		getOpts:   getOpts,
-		watchOpts: watcherOpts,
+		keysAPI:       keysAPI,
+		key:           key,
+		getOpts:       getOpts,
+		watchOpts:     watcherOpts,
+		refreshTicker: refreshTicker,
 	}
 }
 
@@ -30,10 +33,18 @@ type retryWatcher struct {
 	getOpts   *etcd.GetOptions
 	watchOpts *etcd.WatcherOptions
 
+	refreshTicker <-chan time.Time
+
 	cur etcd.Watcher
 }
 
 func (w *retryWatcher) Next(ctx context.Context) (*etcd.Response, error) {
+	// Periodically force a full tree refresh.
+	select {
+	case <-w.refreshTicker:
+		w.cur = nil
+	default:
+	}
 	if w.cur != nil {
 		r, err := w.cur.Next(ctx)
 
@@ -44,7 +55,6 @@ func (w *retryWatcher) Next(ctx context.Context) (*etcd.Response, error) {
 				etcdErr.Code == etcd.ErrorCodeWatcherCleared {
 				w.cur = nil
 			}
-			return w.Next(ctx) // Retry.
 		}
 		return r, err
 	}
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,13 +3,44 @@
 
 [[projects]]
   name = "cloud.google.com/go"
-  packages = ["compute/metadata","internal","internal/optional","storage"]
+  packages = [
+    "compute/metadata",
+    "internal",
+    "internal/optional",
+    "storage"
+  ]
   revision = "8c2dc6124f184bbfc2f13949476cb98e12bda2bb"
   version = "v0.4.0"
 
 [[projects]]
   name = "github.com/aws/aws-sdk-go"
-  packages = ["aws","aws/awserr","aws/awsutil","aws/client","aws/client/metadata","aws/corehandlers","aws/credentials","aws/credentials/ec2rolecreds","aws/credentials/endpointcreds","aws/credentials/stscreds","aws/defaults","aws/ec2metadata","aws/endpoints","aws/request","aws/session","aws/signer/v4","internal/shareddefaults","private/protocol","private/protocol/query","private/protocol/query/queryutil","private/protocol/rest","private/protocol/restxml","private/protocol/xml/xmlutil","service/s3","service/sts"]
+  packages = [
+    "aws",
+    "aws/awserr",
+    "aws/awsutil",
+    "aws/client",
+    "aws/client/metadata",
+    "aws/corehandlers",
+    "aws/credentials",
+    "aws/credentials/ec2rolecreds",
+    "aws/credentials/endpointcreds",
+    "aws/credentials/stscreds",
+    "aws/defaults",
+    "aws/ec2metadata",
+    "aws/endpoints",
+    "aws/request",
+    "aws/session",
+    "aws/signer/v4",
+    "internal/shareddefaults",
+    "private/protocol",
+    "private/protocol/query",
+    "private/protocol/query/queryutil",
+    "private/protocol/rest",
+    "private/protocol/restxml",
+    "private/protocol/xml/xmlutil",
+    "service/s3",
+    "service/sts"
+  ]
   revision = "b9e9981b543687c6fbfd064eaa0f7b69f5fdefd7"
   version = "v1.12.23"
 
@@ -46,13 +77,23 @@
 [[projects]]
   branch = "master"
   name = "github.com/cockroachdb/cockroach"
-  packages = ["util","util/duration","util/encoding"]
+  packages = [
+    "util",
+    "util/duration",
+    "util/encoding"
+  ]
   revision = "0b84c441f9d6f2f6e8f915ea31e8e8978cba63a6"
   source = "https://github.com/jgraettinger/cockroach-encoding.git"
 
 [[projects]]
   name = "github.com/coreos/etcd"
-  packages = ["client","error","pkg/pathutil","pkg/types","store"]
+  packages = [
+    "client",
+    "error",
+    "pkg/pathutil",
+    "pkg/types",
+    "store"
+  ]
   revision = "377f19b0031f9c0aafe2aec28b6f9019311f52f9"
 
 [[projects]]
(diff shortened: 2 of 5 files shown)


pkg/consensus/retry_watcher.go
```diff
@@ -30,16 +34,29 @@ type retryWatcher struct {
...
+	refreshPeriod time.Duration
+	refreshNext   time.Time
+	now           func() time.Time
+
 	cur etcd.Watcher
 }
 
 func (w *retryWatcher) Next(ctx context.Context) (*etcd.Response, error) {
+	// Periodically force a full tree refresh.
+	if w.now().After(w.refreshNext) &&
```
jgraettinger: nit: unnecessary newline, and the `refreshNext.IsZero` & `refreshPeriod != 0` checks don't appear needed, right? isn't it enough to be after `refreshNext`?

With a channel strategy, you'd do a non-blocking select here and iff it selects, nil `w.cur`.

pkg/consensus/retry_watcher.go
```diff
@@ -30,16 +34,29 @@ type retryWatcher struct {
 	getOpts   *etcd.GetOptions
 	watchOpts *etcd.WatcherOptions
 
+	refreshPeriod time.Duration
```
jgraettinger: optional: seems like you could more succinctly use a `time.Ticker` here, with a non-blocking select below.

In fact, I might just pass in a `<-chan time.Time` to `RetryWatcher` and have the caller own the ticker itself (so that it can properly `Stop` it). Note that a nil channel would be perfectly valid and would never select.

For the test below, I'd just make my own buffered channel and feed it a `time.Time{}` at the right point (which also makes it a 100% deterministic test, which technically the test below isn't since there's no guarantee the 10 minutes period doesn't elapse early).

I generally try to avoid lots of `time.Now()` calls -- they're pretty cheap, but they're still system calls. I don't expect it to be a performance problem here though.

pkg/consensus/retry_watcher.go
```diff
@@ -30,16 +34,29 @@ type retryWatcher struct {
...
 
 		if etcdErr, ok := err.(etcd.Error); ok {
 			// If the error code indicates that further Next() attempts will fail,
 			// clear the current etcd.Watcher to force a full tree refresh.
+			// Ignore all other errors. For silent errors, upper-bound the
+			// effects with |refreshPeriod| by defensively forcing
+			// a full tree refresh periodically.
 			if etcdErr.Code == etcd.ErrorCodeEventIndexCleared ||
 				etcdErr.Code == etcd.ErrorCodeWatcherCleared {
 				w.cur = nil
```
jgraettinger: Not your code, but the recursive call to `w.Next` below doesn't seem like a great idea to me, vis a vis just returning the error like any other. That removes the possibility of silent errors; `Allocator` will log a warning and retry `Next` itself.

jgraettinger: (FWIW, I don't think there actually are any other silent errors; the other Etcd errors are driven by bad user input and we can expect to see a stack overflow, but I could be wrong here).

julianvmodesto:
Ack sorry I dropped off on this, PTAL.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/consensus/retry_watcher.go, line 37 at r1](https://reviewable.io:443/reviews/liveramp/gazette/40#-L481Qtp2t6DnRJoW4gH-r1-37:-L4afuDIbNb-Li9l5kwy:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/a824a24e849b6a9330a691430c8be54578d43cae/pkg/consensus/retry_watcher.go#L37)):*

optional: seems like you could more succinctly use a `time.Ticker` here, with a non-blocking select below.

In fact, I might just pass in a `<-chan time.Time` to `RetryWatcher` and have the caller own the ticker itself (so that it can properly `Stop` it). Note that a nil channel would be perfectly valid and would never select.

For the test below, I'd just make my own buffered channel and feed it a `time.Time{}` at the right point (which also makes it a 100% deterministic test, which technically the test below isn't since there's no guarantee the 10 minutes period doesn't elapse early).

I generally try to avoid lots of `time.Now()` calls -- they're pretty cheap, but they're still system calls. I don't expect it to be a performance problem here though.
</blockquote></details>

Done.

---

*[pkg/consensus/retry_watcher.go, line 46 at r1](https://reviewable.io:443/reviews/liveramp/gazette/40#-L481Qtp2t6DnRJoW4gH-r1-46:-L4afyzHj_hiyu8PiWrn:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/a824a24e849b6a9330a691430c8be54578d43cae/pkg/consensus/retry_watcher.go#L46)):*

nit: unnecessary newline, and the `refreshNext.IsZero` & `refreshPeriod != 0` checks don't appear needed, right? isn't it enough to be after `refreshNext`?

With a channel strategy, you'd do a non-blocking select here and iff it selects, nil `w.cur`.
</blockquote></details>

Done.

---

*[pkg/consensus/retry_watcher.go, line 62 at r1](https://reviewable.io:443/reviews/liveramp/gazette/40#-L481Qtp2t6DnRJoW4gH-r1-62:-L4ag5kkQ052jtQT5qCT:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/a824a24e849b6a9330a691430c8be54578d43cae/pkg/consensus/retry_watcher.go#L62)):*

(FWIW, I don't think there actually are any other silent errors; the other Etcd errors are driven by bad user input and we can expect to see a stack overflow, but I could be wrong here).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40#-:-L4ajN7502vW6gHMcJYI:b-ha6nwt)*

jgraettinger:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/consensus/allocator.go, line 105 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4akowsO_6ZhLNrl_x5:-L4akowsO_6ZhLNrl_x6:b-qtav7b) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/allocator.go#L105)):*
> ```Go
> 	defer func() { close(cancelWatch) }()
> 	var refreshTicker = time.NewTicker(time.Minute * 10)
> 	defer func() { refreshTicker.Stop() }()
> ```

nit: fold this into the same `defer` that closes `cancelWatch`.

---

*[pkg/consensus/retry_watcher.go, line 62 at r1](https://reviewable.io:443/reviews/liveramp/gazette/40#-L481Qtp2t6DnRJoW4gH-r1-62:-L4alpJwKaZrNXE_pFcY:b-keujnm) ([raw file](https://github.com/liveramp/gazette/blob/a824a24e849b6a9330a691430c8be54578d43cae/pkg/consensus/retry_watcher.go#L62)):*
<details><summary><i>Previously, julianvmodesto (Julian V. Modesto) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Update comment please (there are no longer any silent errors). 

---

*[pkg/consensus/retry_watcher.go, line 61 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4am2To_2WeKDQ1vH9b:-L4am2To_2WeKDQ1vH9c:b-91ikbe) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher.go#L61)):*
> ```Go
> 				w.cur = nil
> 			}
> 			return r, err
> ```

nit: unnecessary extra return (vs just letting it fall through).

---

*[pkg/consensus/retry_watcher_test.go, line 60 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4am_HIpUkQLBRLJILE:-L4am_HIpUkQLBRLJILF:btxmhg0) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher_test.go#L60)):*
> ```Go
> 		etcd.Error{Code: etcd.ErrorCodeEventIndexCleared}).Once()
> 
> 	resp, err = rwatcher.Next(ctx)
> ```

I don't think `refreshTicker` is actually firing in this test (it's not required that the send be in a goroutine - it's a buffered channel, you can send immediately without a ready receiver, and probably the goroutine is not even being started until after the test has already completed).

Rather, you're observing the fact that the ErrorCodeEventIndexCleared above has already force a full watch reload.

Try reverting all changes to this test ...

---

*[pkg/consensus/retry_watcher_test.go, line 81 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4an0TFtMttPpX1DcBq:-L4an0TFtMttPpX1DcBr:bmimirj) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher_test.go#L81)):*
> ```Go
> 	c.Check(resp.Node, gc.DeepEquals, &etcd.Node{Key: "/key", Value: "four"})
> 	c.Check(err, gc.IsNil)
> ```

... and _here_, doing:

```
refreshTicker <- time.Time{}


keys.On("Get", ctx ...) // Set up another expectation of a full Get & Watcher being done.
keys.On("Watcher", "/key", ...)

resp, err = rwatcher.Next(ctx) // etc.
```

Grab me if this isn't clear to go over in person.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40)*

julianvmodesto:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/consensus/allocator.go, line 105 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4akowsO_6ZhLNrl_x5:-L4i4iOF_ePRWuSg2UOL:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/allocator.go#L105)):*

nit: fold this into the same `defer` that closes `cancelWatch`.
</blockquote></details>

Done.

---

*[pkg/consensus/retry_watcher.go, line 62 at r1](https://reviewable.io:443/reviews/liveramp/gazette/40#-L481Qtp2t6DnRJoW4gH-r1-62:-L4i4wScOnUYux4lcC0-:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/a824a24e849b6a9330a691430c8be54578d43cae/pkg/consensus/retry_watcher.go#L62)):*

Update comment please (there are no longer any silent errors). 
</blockquote></details>

Done.

---

*[pkg/consensus/retry_watcher.go, line 61 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4am2To_2WeKDQ1vH9b:-L4i53CFfS9DZqiV3dz3:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher.go#L61)):*

nit: unnecessary extra return (vs just letting it fall through).
</blockquote></details>

Done.

---

*[pkg/consensus/retry_watcher_test.go, line 81 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4an0TFtMttPpX1DcBq:-L4i5AT6rjj0Yt4TSt7e:b-iryh8y) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher_test.go#L81)):*

... and _here_, doing:

```
refreshTicker <- time.Time{}


keys.On("Get", ctx ...) // Set up another expectation of a full Get & Watcher being done.
keys.On("Watcher", "/key", ...)

resp, err = rwatcher.Next(ctx) // etc.
```

Grab me if this isn't clear to go over in person.
</blockquote></details>

Ohhh I see this makes sense, whoops! Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40)*

julianvmodesto:
PTAL

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40#-:-L4mVhMMB-n-6BxVH-re:bb71pro)*

jgraettinger:
:lgtm_strong: 

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/consensus/retry_watcher_test.go, line 81 at r3](https://reviewable.io:443/reviews/liveramp/gazette/40#-L4an0TFtMttPpX1DcBq:-L4nOnw3H1rs0o-RUR67:b6lqmcl) ([raw file](https://github.com/liveramp/gazette/blob/9f652017d3e76f4bb85252823a4b7851e1331acb/pkg/consensus/retry_watcher_test.go#L81)):*
<details><summary><i>Previously, julianvmodesto (Julian V. Modesto) wrote…</i></summary><blockquote>

Ohhh I see this makes sense, whoops! Done.
</blockquote></details>

looks great!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40#-:-L4nOyzltaTIa7HAF6ul:b-uwkowd)*

jgraettinger:
Reviewed 1 of 4 files at r2, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/40)*

PR #54, opened 2018-03-23
Add error check to Close in SFTP CopyAtomic

tylercunnion:
Currently if a copy is successful, but an error occurs while closing the
file, that error is discarded. We should have visibility on errors which
occur during the close.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/54)

### DIFF
--- a/pkg/cloudstore/filesystem.go
+++ b/pkg/cloudstore/filesystem.go
@@ -42,7 +42,7 @@ type FileSystem interface {
 	// (either via |to.Write()| or *|from.Read()|*), the partially-written content
 	// is removed or never made observable on the target FileSystem (depending on
 	// provider semantics). Otherwise, |to| is visible on the FileSystem after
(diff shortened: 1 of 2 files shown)


tylercunnion:
The context here is that during a cfs-sync to LR sftp servers, we see that files are being transferred completely but no close command is issued and therefore no STOR event is recorded on the server. I'd like to know if something is happening (or more precisely, isn't happening) during the Close().

jgraettinger:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/54#-:-L8KljcjzvSbxsoWgwcM:bnfp4nl)*

jgraettinger:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/liveramp/gazette/54)*

PR #62, opened 2018-06-25
Consistent etcd config

evanryanlin:
We prepend "http://" to the etcd endpoint in gazette itself: https://github.com/LiveRamp/gazette/blob/master/cmd/gazette/main.go#L76

We should handle these endpoints consistently in our command line tools so that users don't need to pass configs in different formats.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/62)

### DIFF
--- a/cmd/gazctl/cmd/root.go
+++ b/cmd/gazctl/cmd/root.go
@@ -81,7 +81,7 @@ func etcdClient() etcd.Client {
 		}
 
 		var err error
-		lazyEtcdClient, err = etcd.New(etcd.Config{Endpoints: []string{ep}})
+		lazyEtcdClient, err = etcd.New(etcd.Config{Endpoints: []string{"http://" + ep}})
 		if err != nil {
 			log.WithField("err", err).Fatal("building etcd client")
 		}


jgraettinger:
Can we flip this, that users are instead required to provide full URLs in all cases? I think we should consider endpoint configuration / flags _not_ in URL form to be deprecated.

Philosophy is, unless there's a particular need otherwise, everything it always a URL. Rationale:
 - it makes the specific service at the endpoint self-documenting via scheme (etcd is behind this IP & port, vs some other service).
 - URL query arguments provide an extensible way to pass along metadata ("here's the file location where you can find service account credentials / an SSL client cert, etc").
 - It allows distinguishing protocol versions, eg "etcd://" vs "etcdv3://".

A reason to consider not doing this right away is that it may create migration steps for downstream users, so we might want to bundle this with other changes.

---

Review status: 0 of 1 files reviewed, all discussions resolved (waiting on @jgraettinger)

---



*Comments from [Reviewable](https://reviewable.io/reviews/liveramp/gazette/62#-:-LFwmEcHE8gIgcOf5cVN:b-e8qzfc)*

evanryanlin:
That seems reasonable to me, although in this particular case the etcd client will always have to be initialized with the http protocol, at least as far as I'm seeing. I think there's an argument for making the behavior consistent within gazette even if we reverse both places where we initialize etcd clients later to require fully URLs, though. Anyone who is using gazctl is already using gazette, so anyone who will be migrating this config will be migrating their gazette config anyway. And in the meantime, having a different format for configs across different parts of gazette may cause confusion.

---

Review status: 0 of 1 files reviewed, all discussions resolved (waiting on @jgraettinger)

---



*Comments from [Reviewable](https://reviewable.io/reviews/liveramp/gazette/62#-:-LFx0KA50vOfeMqKyaci:bldpcla)*

PR #64, opened 2018-07-09
Merge values with the same key in shard_compose

evanryanlin:
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/64)


cmd/gazctl/cmd/shard_compose_test.go
```diff
@@ -124,19 +130,19 @@ func (s *ShardComposeSuite) TestHeapCases(c *gc.C) {
...
 			fn: newHeapIterFunc(
+				First,
 				hexIter(""+
 					"0xaabb 0x2222\n"+
 					"0xfoobar bad value\n"),
 				hexIter(""+
 					"0xaaaa 0x1111\n"+
 					"0xeeff 0x4444\n"),
 			),
-			expectKeys: [][]byte{{0xaa, 0xaa}, {0xaa, 0xbb}},
```
evanryanlin: Note that the expected behavior changes slightly in this case. Since we are merging all values with the same key, we have to be slightly less tolerant of errors. We only know that we've exhausted a key and can safely return a value for that key when we have seen a key that's greater. That means that when hit an error, we can't know that we can reliably return a value for the key that we were processing when we hit the error. In practice, this means that iteration may terminate one step earlier than before.

PR #84, opened 2018-09-05
Repartition mapper

evanryanlin:
@jgraettinger 

I thought I had PRed the shard_split command already, but I guess I only PRed the changes to the shard_compose. This will need some reorganization when gazette 2 fully hits, but here's the shard_split command I used for repartitioning distributions-hashes. I also added a short doc describing how the process of repartitioning works. I'm not sure where the best place is to put it, but I'd imagine there will be other "special case" operations that we want to be able to give some guidance for.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/84)

### DIFF
--- a/cmd/Special_operations.md
+++ b/cmd/Special_operations.md
@@ -1,6 +1,5 @@
-Special Operations
-
 Shard repartitioning
+====================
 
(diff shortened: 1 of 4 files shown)


rupertchen:
Closing this in favor of a different approach built on Gazette v2.

PR #124, opened 2018-12-18
Publish images of tagged releases

rupertchen:
A new CircleCI job builds and pushes liveramp/gazette and
liveramp/gazette-examples images to Docker Hub whenever a tag starting
with "v" is pushed to GitHub. The new images are tagged in Docker with
"latest" and the git tag name, minus the leading "v". The "v" is
stripped because it is conventional for release tags on GitHub to have
the prefix while it is not on Docker Hub.

Additional Notes:

The file extensions were removed from executable scripts following the
recommendations in the Google Shell Style Guide.

The project directory was added as an explicit argument to the
executables to avoid problematic edge cases with deducing the path to
the current script.

Connects #115

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/124)


rupertchen:
@jskelcy: fyi.

Issue #99, opened 2018-11-16
gazctl: journals / shards edit


jgraettinger:
As operator, i should be able to easily edit journals or shards which match a selector, and have them apply when saved or re-open my editor on failure.

Only items which actually changed should be updated in Etcd.

PR #132, opened 2019-01-18
Implement "gazctl {journals,shards} edit"

rupertchen:
I recommend combining commits for review.

In addition to enhancing `gazctl`, I added CircleCI branch filters so branches that start with "draft-" or "draft/" do not trigger builds. This allows developers a way to push partial work to GH without flooding their inbox with build failure notifications.

Connects #99

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/132)


jskelcy:
___
*[v2/cmd/gazctl/shards_list.go, line 118 at r1](https://reviewable.io/reviews/liveramp/gazette/132#-LWrgU6oC1zeaG5_BI7i:-LWrgU6oC1zeaG5_BI7j:b-mjiqgq) ([raw file](https://github.com/liveramp/gazette/blob/5051035506c8e63191f4e09d53ea9280ef79c97b/v2/cmd/gazctl/shards_list.go#L118)):*
> ```Go
> }
> 
> func (cmd *cmdShardsList) outputYAML(resp *consumer.ListResponse) {
> ```

NIT: given that we don't let `os.Stdout` be overwritten could we just remove this function?

jskelcy:
___
*[v2/cmd/gazctl/editor/editor/editor.go, line 91 at r1](https://reviewable.io/reviews/liveramp/gazette/132#-LWvhEoA2YuA6_LFQSOi:-LWvtxLbBP_-X9j2rvmq:bsn2q3i) ([raw file](https://github.com/liveramp/gazette/blob/5051035506c8e63191f4e09d53ea9280ef79c97b/v2/cmd/gazctl/editor/editor/editor.go#L91)):*
<details><summary><i>Previously, jskelcy (Jake Skelcy) wrote…</i></summary><blockquote>

It seems like in some places we are using `log.Error` and in other places, we are writing to `os.Stderr` was there some logic you applied to when to use which?
</blockquote></details>

As a quick follow up I was hoping you could configure log levels to log to different writers. There is a task on logrus so that you can say "log.Error goes to stderr and log.Info goes to stdout" but it is not getting much support from the maintainer.

jskelcy:
___
*[v2/cmd/gazctl/journals_edit.go, line 47 at r2](https://reviewable.io/reviews/liveramp/gazette/132#-LXKAH6wD4wVMscC7SM-:-LXKAH6wD4wVMscC7SM0:b-jftg3g) ([raw file](https://github.com/liveramp/gazette/blob/86404872ba85b503a7418f75f9f92f699f7db939/v2/cmd/gazctl/journals_edit.go#L47)):*
> ```Go
> 
> 	var req = newJournalSpecApplyRequest(&tree)
> 	if err := req.Validate(); err != nil {
> ```

NIT: Perhaps validate in the `newJournalSpecApplyRequest` function.

jskelcy:
Looks good. Left a few nits.

jskelcy:
___
*[v2/cmd/gazctl/editor/editor.go, line 144 at r2](https://reviewable.io/reviews/liveramp/gazette/132#-LXKDswZDeippj_Ms8A5:-LXLhscc-zTmfXJrTMwA:bxlw2rd) ([raw file](https://github.com/liveramp/gazette/blob/86404872ba85b503a7418f75f9f92f699f7db939/v2/cmd/gazctl/editor/editor.go#L144)):*
<details><summary><i>Previously, jskelcy (Jake Skelcy) wrote…</i></summary><blockquote>

It is ok to swallow this error?
</blockquote></details>

Followed up and this only errors so if you use SetDeadLine and we never use that so we dont need to handle the error here.

PR #143, opened 2019-02-01
Create document to outline cost considerations for operating Gazette in the cloud

elynnyap:
This doc has the broader title of "Operational Considerations", but it really focuses on factors affecting costs when using Gazette in the cloud. If there are additional tips on this front that would be great to learn about + add to the doc as that will also help us as we migrate from v1 to v2.

One thing that I wasn't sure if I should add was the `fragment.retention` property from the v2 `JournalSpec`, as I'm not clear on how it works. What initiates the pruning of the fragment once the retention interval has passed? Is this an alternative to the approach of configuring a retention/lifecycle policy through the specific cloud storage provider (with the difference being that it doesn't allow for an intermediate transition to cold storage)?

https://github.com/jgraettinger/gazette/blob/2dd24e48a11dc62380f93c5fb2316e2e626c6d79/v2/test/journalspace.yaml#L43-L45

cc @pintohutch

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/143)


zzz6519003:
looks like a party!

PR #151, opened 2019-02-08
collection of changes from testing reporting

jgraettinger:
Please review commit-by-commit.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/151)


jskelcy:
___
*[v2/pkg/protocol/journalspace/node_test.go, line 352 at r4](https://reviewable.io/reviews/liveramp/gazette/151#-LYDQILT1IJCj5BEMWez:-LYDh0Wd62krdsOwHV-M:brxp9sq) ([raw file](https://github.com/liveramp/gazette/blob/984cb9ff91925a3647a3af9603c5dbcc2008a947/v2/pkg/protocol/journalspace/node_test.go#L352)):*

starting doing this, but realized there are usages where we don't PushDown first in tests. Okay with it as-is? 
</blockquote></details>

Sounds good.

Issue #112, opened 2018-12-05
wire up RocksDB compaction TTL


jgraettinger:
We're now using a version of RocksDB which supports compaction TTLs, but have no way to turn this on from the Go side. It may require adding appropriate `cgo` extensions.

rupertchen:
It appears that `rocksdb` has exposed all the C bindings that we will need. We need to update `gorocksdb` to expose `C.rocksdb_set_options` and passing `ttl` as the key.

Links into rocksdb source:
- https://github.com/facebook/rocksdb/blob/e1de88c8c74bbfbbfc80a8992bd6bb489bdfd194/include/rocksdb/c.h#L728
- https://github.com/facebook/rocksdb/blob/e1de88c8c74bbfbbfc80a8992bd6bb489bdfd194/include/rocksdb/advanced_options.h#L641
- https://github.com/facebook/rocksdb/blob/e1de88c8c74bbfbbfc80a8992bd6bb489bdfd194/include/rocksdb/c.h#L728
- https://github.com/facebook/rocksdb/blob/e1de88c8c74bbfbbfc80a8992bd6bb489bdfd194/options/options_helper.cc#L1893
- https://github.com/facebook/rocksdb/blob/5d4fddfa52768952e58abae5044174ad0ce30f7d/db/db_impl.cc#L714
- https://github.com/facebook/rocksdb/blob/5d4fddfa52768952e58abae5044174ad0ce30f7d/db/column_family.cc#L1142

jgraettinger:
Excellent! Procrastination succeeds again.

rupertchen:
My plan is to first make this change on top of our `update_rocksdb` fork, update gazette to use it, then cherry-pick that as a basis for a PR upstream. After that PR is merged, I'll revisit rebasing our fork.

PR #160, opened 2019-02-20
Update RocksDB from v5.14.2 to v5.17.2

rupertchen:
Ultimately we want to set the TTL for level compaction. This option was
immutable in v5.14.2 and made mutable in v5.15.x. Since we can only set
this option dynamically with the C bindings (after opening the DB), we
must upgrade our version of RocksDB.

Connects #112

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/160)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -695,12 +695,13 @@
   version = "v1.3.0"
 
 [[projects]]
-  digest = "1:6ef99b772e9252c54e362af989aa0ad6be34f304c04e9d52c8431d94b4e2fe2d"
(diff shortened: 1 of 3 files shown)


rupertchen:
I forgot that I had not gotten LiveRamp/gorocksdb#5 merged and the dep here updated yet. I will add another commit to this PR once that happens and then re-add the reviewers for this PR.

rupertchen:
The commits can be reviewed as one. I plan to merge squash this PR.

PR #164, opened 2019-02-26
grab bag of updates

jgraettinger:
* Make the journal `flag` field easier to work with in YAML.
* Rework `extraArgs` of the consumer chart. Makes it waaay easier to pass further specializations in from application charts.
* Sign fragments only if requested.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/164)


jskelcy:
___
*[v2/charts/consumer/templates/deployment.yaml, line 56 at r2](https://reviewable.io/reviews/liveramp/gazette/164#-LZee6ZY3eVvXpGq2OHL:-LZee6ZY3eVvXpGq2OHM:bbrtob3) ([raw file](https://github.com/liveramp/gazette/blob/dce5944b5ae29e922802f9c5b99828d886ea33fe/v2/charts/consumer/templates/deployment.yaml#L56)):*
> ```YAML
> {{- end }}
>                 --broker.address {{ or .Values.global.gazette.endpoint .Values.gazette.endpoint }} \
>                 --consumer.port  8080 \
> ```

It would appear that we setting dynamic values in the template and static values in the `values.yaml` file. If that is the case can should `--consumer.port` also be moved?

Issue #163, opened 2019-02-24
client: context cancellation should propagate to AsyncAppends


jgraettinger:
There's a rare race case we're sometimes hitting in tests where a RocksDB sync call / write-barrier blocks forever on test tear-down, because its corresponding recoverylog journal has already been cancelled / torn down.

`client.AsyncAppend` should have an `Err() error` function which pairs with its existing `Done() <-chan struct{}`. It should return `context` errors only (`Cancelled` or `DeadlineExceeded`), and _exclusively_ because the parent `Context` of the `AppendService` was cancelled; everything else is retried.

The `AppendService` should be wired up to propagate context errors to all outstanding `AsyncAppends`.

PR #168, opened 2019-03-01
client: AppendService aborts AsyncAppends when context is cancelled.

jgraettinger:
Update all call-sites of AsyncAppend.Done with appropriate handling or
plumbing for AsyncAppend.Err.

consumer.NewReplica now creates an AppendService using the replica context,
and not context.Background, with the implication that all outstanding appends
are aborted immediately along with the replica.

Issue #163

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/168)


jgraettinger:
FYI I've confirmed this PR fixes occasional internal test hang flakes we were seeing on internal repos.

PR #190, opened 2019-03-22
Replace conditionals with polymorphism

rupertchen:
By introducing the `fragment.backend` interface and refactoring the
package-level functions, the many similar switch statements in
`stores.go` were removed. Furthermore, the metrics instrumentation in
each of the backend implementations were consolidated (moved "up") to
the operation and no longer require deferring a closure.

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/190)

### DIFF
--- a/v2/pkg/fragment/store_fs.go
+++ b/v2/pkg/fragment/store_fs.go
@@ -1,11 +1,13 @@
 package fragment
 
 import (
+	"context"
 	"io"
 	"io/ioutil"
 	"net/url"
 	"os"
 	"path/filepath"
+	"time"
 
 	pb "github.com/LiveRamp/gazette/v2/pkg/protocol"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


jskelcy:
___
*[v2/pkg/fragment/store_gcs.go, line 27 at r1](https://reviewable.io/reviews/liveramp/gazette/190#-LapCXs0ACBN7SCDIO3f:-LapCXs0ACBN7SCDIO3g:b-pf79hv) ([raw file](https://github.com/liveramp/gazette/blob/fcfc05b5d79fe21f9c48361236528df139b21dac/v2/pkg/fragment/store_gcs.go#L27)):*
> ```Go
> 
> type gcsBackend struct {
> 	sharedGCS struct {
> ```

any reason all the properties on `gcsBackend.sharedGCS` could not be moved to be directly on the `gcsBackend` struct?

Issue #194, opened 2019-04-02
broker emitting noisy warnings with GCS backing store


rupertchen:
# Problem
There are a large number of warnings frequently being logged by brokers assigned to journals with stores on GCS. There is a warning from each journal every few minutes, but across all journals, this adds up to a lot of noise.

That said, this issue does not prevent the broker from performing its duty, so it is merely noisy, not broken.

# Proposals
The hypothesis is that the containing directory is included in the iteration across the objects since it matches the prefix in the query.

- Look for a way to exclude the directory from the results.
- Recognize the directory do not attempt to parse its name as a fragment file.
- Suppress the warning when the object is the directory.

# Example logs
```
Apr 2 11:59:45 hoping-pike-gazette-ffb6f6479-qclbr gazette warning {"bucket":"XXXX","err":"wrong Fragment format: YYYY","level":"warning","msg":"parsing fragment","name":"ZZZZ/YYYY/","time":"2019-04-02T15:59:45Z"}
```

jgraettinger:
What's odd about this is that _something else_ is creating "directory"-like empty objects in the fragment store. gazette itself doesn't do this. Still, handling it is reasonable, as there are legitimate reasons for wanting these (eg, it's being mounted / interacted with as a GCS FUSE file-system, for example).

jgraettinger:
Fixed.

PR #195, opened 2019-04-03
Remove extraneous warnings from GCS backed stores

rupertchen:
With this change, gcsBackend#List ignores objects that represent
directories when iterating through the objects of a bucket for a
journal. Such objects never represent fragment files and attempting to
parse their names as such will almost certainly fail, writing a warning
to the logs when, in fact, everything is as expected.

Connects: #194

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/195)


rupertchen:
Remaking pull request with different branch name to trigger CI.

Issue #198, opened 2019-04-20
broker: unexpected FRAGMENT_MISMATCH


jgraettinger:
There's a bug in `pipeline.go:synchronize()`, where it treats a FRAGMENT_MISMATCH status from a replication peer as unexpected when the primary spool is `Begin: 0, End: 0`, and the replication peer is `Begin: 0, End: >0`. The current check requires that `Begin` be different (this is wrong).

jgraettinger:
Fixed.

PR #199, opened 2019-04-21
task.Group refactor & other fixes

### LINKED ISSUE #197: broker: defer persister.Finish()
An Etcd partition caused lease revocations of brokers, and they panic'd (trace below) before they completed uploads of remaining open spools.

This particular error should arguably be handled explicitly, but more broadly brokers should do everything possible to ensure spools are uploaded even if a top-level panic occurs. Specifically, `persister.Finish()` should be defer'd or recover'd in some way.

One other realization from this incident is that brokers should immediately roll the spool after the very first commit (starting byte zero) of a journal, to get a fragment in the backing store as soon as possible. The reason is that this "dirties" the fragment store even if a journal is very low volume and, if an Etcd split were to later occur, it ensures that future allocated brokers don't mistake it for a journal which has never before been written.

```
{"err":"member key not found in Etcd","level":"panic","msg":"Allocate failed","time":"2019-04-15T08:18:52Z"}
panic: (*logrus.Entry) (0xff9920,0xc00b807860) [recovered]
        panic: (*logrus.Entry) (0xff9920,0xc00b807860)

goroutine 1 [running]:
github.com/LiveRamp/gazette/v2/pkg/mainboilerplate.InitDiagnosticsAndRecover.func2()
        /go/src/github.com/LiveRamp/gazette/v2/pkg/mainboilerplate/diagnostics.go:44 +0x11f
panic(0xff9920, 0xc00b807860)
        /usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus.Entry.log(0xc00009a540, 0xc0046f5bf0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus/entry.go:227 +0x2cf
github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus.(*Entry).Log(0xc00b807800, 0xc000000000, 0xc000ca78a8, 0x1, 0x1)
        /go/src/github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus/entry.go:256 +0xb7
github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus.(*Entry).Panic(0xc00b807800, 0xc000ca78a8, 0x1, 0x1)
        /go/src/github.com/LiveRamp/gazette/vendor/github.com/sirupsen/logrus/entry.go:294 +0x55
github.com/LiveRamp/gazette/v2/pkg/mainboilerplate.Must(0x1155e00, 0xc000950e20, 0x10139a1, 0xf, 0x0, 0x0, 0x0)
        /go/src/github.com/LiveRamp/gazette/v2/pkg/mainboilerplate/diagnostics.go:59 +0x27d
github.com/LiveRamp/gazette/v2/pkg/mainboilerplate.AnnounceServeAndAllocate(0xc000396600, 0xc000384a80, 0xc00009b860, 0xc0003e8000, 0x115e380, 0xc00049bec0)
        /go/src/github.com/LiveRamp/gazette/v2/pkg/mainboilerplate/service.go:78 +0x31f
main.serveBroker.Execute(0xc0001b8270, 0x0, 0xd, 0x0, 0x0)
        /go/src/github.com/LiveRamp/gazette/v2/cmd/gazette/main.go:63 +0x8d5
```

jgraettinger:
This is a collection of patches addressing outstanding issues I'm aware of from recovery during recent net-splits.

Testing:
 * I've done a bunch of flake hunting with unit tests (eg `--count=X00`) and feel pretty confident here.
 * I've run it through integration tests on a local stack. Plan to do more manual testing here, but wanted to kick this over now.

Fixes #197, #198

---
This change is [Reviewable](https://reviewable.io/reviews/liveramp/gazette/199)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -832,6 +832,14 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:5e9d1a4b8508ec7da2aca6a46b1dc45dda3a7d8699f152d53fde93646c6cedf1"
+  name = "golang.org/x/sync"
+  packages = ["errgroup"]
+  pruneopts = ""
+  revision = "56d357773e8497dfd526f0727e187720d1093757"
+
+[[projects]]
+  branch = "master"
   digest = "1:0f39f8ca1c492b3b59020cb0f95e6635d5051bf7048a32ac5f9c6bf1adae6ec8"
   name = "golang.org/x/sys"
   packages = [
@@ -1034,6 +1042,7 @@
     "golang.org/x/net/context",
     "golang.org/x/net/trace",
     "golang.org/x/oauth2/google",
+    "golang.org/x/sync/errgroup",
     "google.golang.org/api/gensupport",
     "google.golang.org/api/googleapi",
     "google.golang.org/api/iterator",
--- a/v2/cmd/gazette/main.go
+++ b/v2/cmd/gazette/main.go
@@ -1,6 +1,11 @@
 package main
 
 import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+
 	"github.com/LiveRamp/gazette/v2/pkg/allocator"
 	"github.com/LiveRamp/gazette/v2/pkg/broker"
 	"github.com/LiveRamp/gazette/v2/pkg/fragment"
@@ -9,6 +14,7 @@ import (
 	"github.com/LiveRamp/gazette/v2/pkg/metrics"
 	"github.com/LiveRamp/gazette/v2/pkg/protocol"
 	"github.com/LiveRamp/gazette/v2/pkg/server"
+	"github.com/LiveRamp/gazette/v2/pkg/task"
 	"github.com/jessevdk/go-flags"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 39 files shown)


jskelcy:
___
*[v2/pkg/task/group.go, line 46 at r1](https://reviewable.io/reviews/liveramp/gazette/199#-Ld4Smjh6JSthYny6vjI:-Ld4Smjh6JSthYny6vjJ:bd4ciou) ([raw file](https://github.com/liveramp/gazette/blob/383a84113229cd30399309dc23382b6349310664/v2/pkg/task/group.go#L46)):*
> ```Go
> // Cannot be called after Start is invoked.
> func (g *Group) Queue(desc string, fn func() error) {
> 	if len(g.tasks) == 0 && cap(g.tasks) != 0 {
> ```

I am not sure this is thread safe, this might be super contrived but let me know what you think:
```
group := NewGroup(ctx)
for t := range tasks{
  group.Queue(t.desc, t.fn)
}
go group.Start()
group.Queue("extra tast", extraTask)
```
^^^ In this case if the tasks are still being kicked off then the final queue will not cause a panic when it should.

Would it be safer to set an atomic boolean on the `Group` struct (maybe `started`) which is set to true when `Start` is called and that can just be checked in the `Queue` method?

jgraettinger:
Start is to be called synchronously, and it returns immediately. Will
document that queue and start are not themselves threadsafe.

jgraettinger:
Maybe a better name is GoRun() or something to make clear it's invoking go

jskelcy:
___
*[v2/pkg/allocator/announce.go, line 120 at r1](https://reviewable.io/reviews/liveramp/gazette/199#-Ld4_H1hDY8iBkN8-owW:-Ld5IRn07IkACmnZWC60:b4c9lyr) ([raw file](https://github.com/liveramp/gazette/blob/383a84113229cd30399309dc23382b6349310664/v2/pkg/allocator/announce.go#L120)):*

Yes. The set of actual signals to monitor is set with the `Notify` call in main().
</blockquote></details>

:thumbsup:

jskelcy:
___
*[v2/pkg/task/group.go, line 66 at r1](https://reviewable.io/reviews/liveramp/gazette/199#-Ld4U5dO1PdXXKqObp4a:-Ld5Nf744zTFzHntf9V9:bi6q7qx) ([raw file](https://github.com/liveramp/gazette/blob/383a84113229cd30399309dc23382b6349310664/v2/pkg/task/group.go#L66)):*

> but it is the case before the tasks can be enqueued the endpoint author needs to check to see if the list is empty and make sure to skip the Group related logic. 

I didn't follow this ^

I did update this bit such that `GoRun` and `Wait` can be called even if `Queue` never is, and to clarify handling re: panics if the Group state-machine is misused.
</blockquote></details>

This looks great. 

The issue was talking about, which you have fixed,  was something like this.
```
g := newGroup(ctx)
userProvidedList := req.List

for  t := range userProvidedList {
  g.Queue(t)
}

g.Start()
g.Wait()
```

^^ basically this code would panic if the request passed in from the user had and emty list, which might be valid for that api. Instead, you would need something like this:
```
userProvidedList := req.List
if len(userProvidedList) >0 { 
  g := newGroup(ctx)
  for  t := range userProvidedList {
    g.Queue(t)
  }

  g.Start()
  g.Wait()
}
```
But with the new logic this is not longer and issue

jskelcy:
___
*[v2/pkg/task/group.go, line 46 at r1](https://reviewable.io/reviews/liveramp/gazette/199#-Ld4Smjh6JSthYny6vjI:-Ld5PUz66xsoJIJBRVUa:b-896fix) ([raw file](https://github.com/liveramp/gazette/blob/383a84113229cd30399309dc23382b6349310664/v2/pkg/task/group.go#L46)):*

Unpacking my thought process further: ideally package `task` wouldn't exist, but:

  * I did want to plumb task definition & queueing down into eg `allocator` & `server`, and other low-level packages which don't otherwise relate.
 * The `errgroup.Group` must have already been created (as it provides the Context that closures should be monitoring), but
 * We don't want to actually start any goroutines yet, as it's easier to reason about if they're only started after all other initialization is fully done, which means _something_ must retain and track the tasks to run. This is thus where I landed.
</blockquote></details>

Done.

jskelcy:
___
*[v2/pkg/allocator/announce.go, line 120 at r1](https://reviewable.io/reviews/liveramp/gazette/199#-Ld4_H1hDY8iBkN8-owW:-Ld5Tjsz74gcpmkMVsu8:b-uybo9b) ([raw file](https://github.com/liveramp/gazette/blob/383a84113229cd30399309dc23382b6349310664/v2/pkg/allocator/announce.go#L120)):*

> Definitely don't want to do that. That cancels everything. We must handle by zeroing and updating the member spec, so that the allocator can do it's thing to move load around.

I think I misread your question. Trying again: The reason this is here and not in main or `mainboilerplate` is that, by having it here, it's now actually used by unit tests building upon `brokertest` and `consumertest`. Those implementations also got to drop their own specialized logic for zeroing their member limit in favor of using this (they create a chan os.Signal and directly send on it, rather than using `Notify`).
</blockquote></details>

I suppose I need to look more into notify but given that this function call comes from main I can see it being ok. 

My gut reaction was that I may want to do a lot of things when I receive a sig, maybe persist something to disk/cloud storage, log something, write some keys to etcd. By handling the signal in main and then explicitly canceling the relevant contexts which are each responsible for doing some "cleanup". I can see now how with the combination of this logic and `task.Group` in main you can still do arbitary cleanup task function with the `desc` of `service.Watch` in main.

jskelcy:
there are currently some stuck consumer shards. I am thinking that the fragment mismatch fixes should get those shards out of the failed state. I can report back when this had been landed and deployed.

Issue #229, opened 2019-11-13
global and per-journal append quotas


jgraettinger:
As a cluster operator, I want to:
 1) Set global bounds on per-journal append quotas
 2) Configure quotas on each JournalSpec (which would be effectively capped by the global setting)
 3) Configure cross-journal quotas on a per-client basis.

... in order to provide overall QoS and better predict broker utilization of a shared-tenancy cluster.

*3* isn't in scope right now, as it requires a better developed authentication/authorization story. It's also unclear how a granular per-client quota would be enforced without delaying appends of other clients which have available quota.

*1 & 2* are in scope. Plan is to:

 * Add (optional) broker flags:
    `--broker.max-append-rate: upper bound on per-journal steady state append rate, in bytes per second (0 is unlimited) [default: 0]`
    **maybe** `--broker.max-burst-factor: upper bound on the allowed variance from max-append-rate, in multiples of max-append-rate [default: 3]` *(for now, this will be hard-coded to keep the API simpler. Let's revisit later if there's demonstrated need for this knob)*
 * Add corresponding `JournalSpec` fields
 * Apply github.com/throttled/throttled to do the heavy lifting (GCRA-based implementation)
    - Computing the effective period and burst factor as min(JournalSpec, global flag).
    - Effect throttling on each client chunk within [append_fsm.run()](https://github.com/gazette/core/blob/master/broker/append_fsm.go#L76-L121).

PR #230, opened 2019-11-16
broker: journal-specific & global max-rate flow control

jgraettinger:
Flow control uses a budgeting / token-bucket strategy. Individual append
chunks which would violate the flow-control budget are held back until
sufficient budget is available to be redeemed.

Min-rate flow control is re-worked from the current binary "saw
chunk"/"didn't see chunk" strategy to actually require a sustained
minimum flow rate.

A new JournalSpec field is added for per-journal control over maximum
rate. Broker flags for minimum rate and a global maximum rate are added.

Issue #229

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/230)


jgraettinger:
@jskelcy  let me know if you don't have bandwidth for this. Thanks

Issue #238, opened 2019-11-18
fragments: support template sub-path prefixes


jgraettinger:
As a solutions architect, I want to directly leverage fragment files written by Gazette in an ad-hoc query environment, via a external Athena/BigQuery/Hive external table, or in my Spark-powered "data lake", or both. A commonality of these environments is their support for (virtual) directory partitioning schemes, like `path/to/dataset/source=web/region=US/date=2019-11-19/hour=11`.

Some of this can be specified in the journal name itself, but it would be silly to create a new journal for every date or even hour. Instead, JournalSpecs should support a templated fragment sub-path under which the fragment will be persisted, like `date={ daystamp .Fragment.StartTime }/hour={ hour .Fragment.StartTime }`. Fragment listings should similarly support walking (virtual) sub-directories.

PR #242, opened 2019-11-20
broker: add JournalSpec.Fragment.PathPostfixTemplate

jgraettinger:
Primary use-case of path postfix templates is to allow for Hive-compatible
partitioning of a journal's fragments on creation time, under the path
already implied by the fragment store prefix & journal name.

Using templates allows for variation in formatting, and allows for other
esoteric use cases (i.e. using journal registers, or current properties
of the JournalSpec).

Issue #238

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/242)


jgraettinger:
(skip past the first commit in review; this PR is rebased upon #230 because they conflict).

PR #245, opened 2019-12-04
consumer: return descriptive error when a nil state is passed into NewJSONFileStore

kanwals:
Return a more descriptive error if a `nil` `state` into `NewJSONFileStore`.

Didn't add/modify tests as the change is pretty trivial. LMK if you think otherwise.


kanwals:
Just a headsup that I'm aware of the response, and would likely need some time (probably EOW) to get to this.

PR #250, opened 2020-01-18
remove etcd dependency from broker/client and broker/protocol

michaelschiff:
Removes etcd dependencies from broker/client and broker/protocol

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/250)


michaelschiff:
Makes sense, will update accordingly

PR #254, opened 2020-02-12
adds grpc-ecosystem go-grpc-prometheus interceptors to server.go

michaelschiff:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/254)


michaelschiff:
Not sure how you feel about the addition of another dependency, but its nice to have the same standard metrics on all grpc services. Another option is giving the user an ability to pass options to the GRPC server (which may actually be more flexible long-term).

michaelschiff:
I agree, that feels like a more flexible approach.

jgraettinger:
Reconsidered this, and believe it's a good idea. Building on the grpc-prometheus ecosystem allows for things like mixing in prepackaged Grafana dashboards / Prometheus rules / alerts etc for exploring these metrics (though to be honest I'm not finding much, beyond https://github.com/grpc-ecosystem/go-grpc-prometheus/issues/68)

PR #257, opened 2020-04-22
Update all the dependencies

jgraettinger:
Best to review this commit-by-commit in Reviewable. High level:

 * Update to Go 1.14.2, and update a bunch of go.mod dependencies.
   - gRPC made a breaking change to the load balancer interface that required updates
   - Rebuild protobufs with newer gogoproto.
 * Dependency on Etcd server was becoming pretty gross. Rip it out, in favor of spawning off an `etcd` subprocess instead that uses the new `testing.Main` API. Gazette depends only on the Etcd client now.
 * Update RocksDB to 6.7.3 (and fix a bug in ArenaIterator).

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/257)


jgraettinger:
Thanks, updated changelog with this as well as some other notable updates not already covered.

PR #260, opened 2020-05-25
address UUID sequencing race

jgraettinger:
The action is in the final commit. Included a few other commits fixing / instrumenting minor test flakes you can skip over if desired. Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/260)


michaelschiff:
https://github.com/gazette/core/blob/4f60807b5c755be6e3cde75f5269d0248afc59d9/message/sequencer.go#L118

How do you feel about adding a metric for this?  Labels something like shardId and source journal.

michaelschiff:
LGTM

PR #268, opened 2020-06-18
Update build docs

psFried:
Small docs update to help other new contributors get up to speed. The more commonly used build commands (using the `as-ci` target) are now closer to the top to emphasize their use over the plain make targets. Information about the plain build targets was moved toward the bottom, after a newly added section about required dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/268)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 .build
 .build-ci
+
+docs/_build
+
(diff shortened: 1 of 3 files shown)


jgraettinger:
👏 first commit down!

PR #272, opened 2020-07-02
migrate docs from circleci to github actions

psFried:
Just some minor docs updates related to the change from circleci to github actions. 

One thing that's _not_ covered here is the image names used in the kustomize templates. These currently use `gazette/broker` (without a registry host). One of the benefits(?) of docker hub is that it's the default registry, so image names without a host will automatically be pulled from there. There's a couple of ways that we could address that in the future, but one idea would be to add a `kustomization.yaml` that just modifies the image names.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/272)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,9 +3,9 @@
    :alt: Gazette Logo
 
 
-.. image:: https://circleci.com/gh/gazette/core.svg?style=svg
-   :target: https://circleci.com/gh/gazette/core
-   :alt: CircleCI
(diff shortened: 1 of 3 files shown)


jgraettinger:
I think we could just update the default `gazette/broker` to point to quay.io.

A little bit down the line, I'm tempted to altogether junk the `kustomize` folder and replace with jsonnet-derived manifests.

Issue #273, opened 2020-07-06
Improve build process for linking to native libraries


psFried:
The current situation for gazette builds has a couple of issues when it comes to linking native libraries that adds a few headaches to building consumer applications. Most of these issues center around rocksdb. Most native library dependencies are simply provided by environment that we're building in. Rocks has been treated differently because gazettte uses features that require "Runtime Type Information", which requires a compile-time flag to enable. This requires building rocksdb from source. Since we still wanted consumers to be able to choose their own rocksdb version, a separate make target was added, allowing users to set `ROCKSDB_VERSION` by passing a make variable. This means that there are different methods for building against specific versions of native libraries, depending on which library we're talking about. Additionally, the way that linker args are specified was making it difficult to control things like static linking.

I'd like to try cleaning this up a bit. The specific proposal is to first move rocksdb out of the makefile and into the ci-builder docker image, thus making the build environment the sole provider of native dependencies. A reasonable next step would be to separate out the code for building and running the ci-builder docker container to create a clear separation between the portions of the gazette build infrastructure that are meant to be shared and those that are not. The ultimate goal is to only share the docker image, and not any of the make targets.

psFried:
Closing this, as the desire is to continue using the shared make file at this point. We can re-open another issue if we decide that this is something we wish to address in the future,

PR #274, opened 2020-07-06
Rocks builds

psFried:
This addresses a few of the issues with building/linking native libraries, and also fixes some issues with the docker-based build environments creating files that can't be deleted. See #273 for more context. This PR addresses the first portion of that, but more work will be needed in order to make it easier to build consumer applications without needing to share the main gazette makefile.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/274)


psFried:
Closing this PR, as the desire is to stick with the shared makefile as that mechanism allows for building the native libs outside of the docker-based environment

PR #293, opened 2021-08-20
Gazette Updates from Flow Development Omnibus 🚌 

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/293)


jgraettinger:
This mega-PR is a feature branch which was forked off from Gazette a bit over a year ago, during the early development of [Estuary Flow](https://github.com/estuary/flow). It was forked to allow for experimentation and false starts along the way, as we figured out Flow, but in practice there were only a couple of things to back out.

@psFried and I reviewed these commits along the way, and I've re-reviewed everything over the past few days, along with cleaning up / backing out / adding a few missing tests.

Qualitatively, the changes are:

 * Allowing applications to provide an alternate driver of message reading & parsing, as opposed to reading journals named in the ShardSpec.
     * Flow uses this to power a read-time data shuffle across a partitioned collection of documents.
 * Updating recovery logs to allow for cheap "forks", in a backwards-compatible way.
     * Flow uses this to implement its online shard-splitting workflow.
 * Allowing journal names in Read and Append RPCs to be extended with a 'name;suffix', where the suffix is captured into consumer checkpoint offsets.
      * This allows a single consumer to have multiple, concurrent reads of a journal.
 * New instrumentation for testing and building custom consumer executables.
 * Fixing a memory leak in append flow control.
 * Fixing an allocator bug which could cause delayed convergence under pretty uncommon conditions.
 * Fixing some corner cases on timely reporting of shard progress in the Stat RPC.
 * Various other minor bugs or annoyances.

allocator/allocator.go
```diff
@@ -318,37 +330,53 @@ func (b *batchedTxn) Checkpoint() error {
...
+	}
+
+	if err != nil {
 		return nil, err
-	} else if !r.Succeeded {
-		return r, fmt.Errorf("transaction checks did not succeed")
+	} else if !response.Succeeded {
+		return response, fmt.Errorf("transaction checks did not succeed")
 	} else {
 		if len(b.ops) > 0 {
```
michaelschiff, line 348: It seems like this check may no longer be necessary? Since we check the length of ops above and exit early it looks like this condition will always be true (unless it is possible for `txnDo` to modify `ops` to be empty, though is this still considered a NOOP?).  If this is the case we could just set `b.noop = false` near the top after the set of checks.

jgraettinger: Yep, you're right, this is vestigial now.

PR #296, opened 2021-09-10
This refactors the gazctl command to allow importing the cli parsing to another project

snowzach:
It had to be reworked a bit because the CLI/Options parser did most of it's work in init functions and to allow importing it from another project required setting up a bunch of register functions so the commands can be built once the program is started.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/296)


jgraettinger (approved):
Nice! Thanks

Issue #297, opened 2021-09-10
client: AppendService doesn't release buffers after cancellation


jgraettinger:
A very busy writer, publishing into an AppendService which has already been cancelled, will quickly produce many append buffers and exhaust the available file limit.

~~This happens (I think) because `serveAppends` quickly races ongoing publishes, grabbing their backing AsyncAppend to mark it as errored with a cancellation. The ongoing publish in turn creates a new AsyncAppend, and so on, resulting in lots of AsyncAppends with backing (small) file buffers.~~

On investigation, turns out that AppendService fails to release `appendBuffer`s if the append is already error'd. For example because the journal was marked with a previously-seen cancellation.

PR #299, opened 2021-09-15
client: AppendService owns its buffer sync.Pool

jgraettinger:
Rather than using a global singleton Pool.

In the future this may allow for better coupling of resource quotas
and back-pressure against specific AppendService instances,
correlated to different consumer.Shards.

The immediate motivation is fixing a test-only bug, where a
mockFile{} fixture is added to the global Pool and causes a
future test case to fail.

Issue #297

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/299)


psFried (approved):
Looks good to me. One thought that occurs is that the lifecycle of each pool is now potentially a lot shorter. Is there anything we should be doing to cleanup the pools when an append service is shutdown, or is it ok to let it be handled by the garbage collector?

jgraettinger:
A-ok for the garbage collector to deal with them. Thanks

PR #311, opened 2021-11-02
multiple improvements to facilitate local testing contexts

jgraettinger:
Thematically this PR is a bag of quality-of-life improvements that make it easier to run local testing clusters.

* Make it possible to run vanilla shard / journal specs in a single-process cluster by bounding effective replication and disabling remote stores
* Support Unix domain sockets so as not to pollute the port space if you're running a lot of these test clusters
* Squelch a spurious warning.

See individual commits.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/311)

### DIFF
--- a/broker/client/reader.go
+++ b/broker/client/reader.go
@@ -364,12 +364,13 @@ func mapGRPCCtxErr(ctx context.Context, err error) error {
 
 var (
 	// Map common broker error statuses into named errors.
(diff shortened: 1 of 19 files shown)


jgraettinger:
I've gone ahead and merged to unblock some downstream work, but review is still welcome and I'll address feedback as follow-ups.

PR #312, opened 2021-12-15
Add `shard unassign` cmd

saterus:
Adds a `shard unassign` command to gazctl. This allows manually triggering the reallocation of a shard to a different consumer process. This provides a quick way to restart a failed shard or to try to relieve the load on a hot node.

I tried to follow the conventions I saw elsewhere in the codebase. Let me know if I've done anything strange or put responsibilities in the wrong place.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/312)


jgraettinger (changes requested):
Looking great, a few comments below. Thanks

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312#-:-Mr089BD8GdEyxT42btb:bv23rie)* status: all files reviewed, 7 unresolved discussions (waiting on @saterus)

___
*[cmd/gazctl/gazctlcmd/gazctl.go, line 74 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr01Ssg4yTzUCr3Vr3o:-Mr01Ssg4yTzUCr3Vr3p:bdqmipa) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/gazctl.go#L74)):*
> ```Go
> }
> 
> type unassignConfig struct {
> ```

nit: I think this can be inlined into `cmdShardsUnassign` (or you could re-use `pruneConfig`?). Other structs in here were all extracted because they're used in multiple places (typically for a journals vs shards flavor of an operation).

for now, or later: What about a boolean Failed which further restricts the selection to primary shards which have a status of `FAILED`, and leaves others alone ?
___
*[cmd/gazctl/gazctlcmd/shards_unassign.go, line 52 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr02Gzm6MFhe6Oq-3Iq:-Mr02Gzm6MFhe6Oq-3Ir:b-mpv44x) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/shards_unassign.go#L52)):*
> ```Go
> 
> 	if len(shards) > 0 {
> 		log.Infof("Successfully unassigned %v shards: %v", len(shards), shardIds(shards))
> ```

supernit: perhaps log a line for each shard? That can also include more information, like its primary status, and its number of route members.
___
*[consumer/shard_api.go, line 197 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04lgE0HXP7gnS6qcL:-Mr04lgE0HXP7gnS6qcM:b-6cjxsc) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L197)):*
> ```Go
> 		return resp, err
> 	}
> ```

You'll need to RLock the KeySpace mutex here, to guard access to Resolver.state
___
*[consumer/shard_api.go, line 204 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04o4UEmXfzz-lgpZy:-Mr04o4UEmXfzz-lgpZz:b-64lv86) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L204)):*
> ```Go
> 	}
> 
> 	etcdResp, err := srv.Etcd.KV.Delete(ctx, allocator.AssignmentKey(state.KS, primaryAssignment))
> ```

If the primary shard has status < FAILED, then this should further check that:
 * The number of assignments matches the desired replication.
 * All shard assignments are consistent.

These are sanity checks which help sure the required replication invariants of the shard.
If the primary is FAILED, we can of course be more willing to remove it.

See:
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/allocator#IsConsistentFn
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/consumer#ShardIsConsistent

Any other potential sanity checks you're aware of?
___
*[consumer/protocol/protocol.proto, line 447 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr06LgI8PGSkQRuKgR0:-Mr06LgI8PGSkQRuKgR1:b-o6vwxq) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/protocol/protocol.proto#L447)):*
> ```protobuf
> 
> message UnassignRequest {
>   // Header may be attached by a proxying consumer peer.
> ```

I think you can drop Header, since this isn't a proxied request (any consumer member can serve it).
___
*[consumer/protocol/protocol.proto, line 450 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr06QHBDCgVwJIGiMbR:-Mr06QHBDCgVwJIGiMbS:b5n1x44) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/protocol/protocol.proto#L450)):*
> ```protobuf
>   protocol.Header header = 1;
>   // Shard to unassign.
>   string shard = 2 [ (gogoproto.casttype) = "ShardID" ];
> ```

Should this be `repeated string`, allowing for a bunch of shards to be assigned as a single RPC ?
___
*[consumer/protocol/rpc_extensions.go, line 157 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr06l3r5o5qHnHavFKN:-Mr06l3r5o5qHnHavFKO:bxktquh) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/protocol/rpc_extensions.go#L157)):*
> ```Go
> // Validate returns an error if the UnassignRequest is not well-formed.
> func (m *UnassignRequest) Validate() error {
> 	if err := m.Shard.Validate(); err != nil {
> ```

Test in https://github.com/gazette/core/blob/master/consumer/protocol/rpc_extensions_test.go ?

Yes, it's very pedantic :) However the exercise has caught my bugs in the past -- here, I'm noting that Header isn't validated (though I suggested it's removal anyway, above).

saterus:
___
*[consumer/protocol/protocol.proto, line 450 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr06QHBDCgVwJIGiMbR:-Mr3LEiP-Vv6wgkJkMdh:bjggcce) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/protocol/protocol.proto#L450)):*

Should this be `repeated string`, allowing for a bunch of shards to be assigned as a single RPC ?
</blockquote></details>

Sure thing. It raises the question of error handling though. Would you expect the assignment removals to be transactional? Or if we encounter an error with one of many, should we just bail?

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312)* status: all files reviewed, 7 unresolved discussions (waiting on @saterus)

___
*[consumer/shard_api.go, line 204 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04o4UEmXfzz-lgpZy:-Mr3NO2m1Ozk9ZuFeRHJ:b-mwgwf2) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L204)):*

If the primary shard has status < FAILED, then this should further check that:
 * The number of assignments matches the desired replication.
 * All shard assignments are consistent.

These are sanity checks which help sure the required replication invariants of the shard.
If the primary is FAILED, we can of course be more willing to remove it.

See:
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/allocator#IsConsistentFn
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/consumer#ShardIsConsistent

Any other potential sanity checks you're aware of?
</blockquote></details>

Oh, one more I have: checking the modification revision of the assignment hasn't changed out from under you within the Etcd operation. I think this could actually matter if there were raced invocations of `gazctl shards unassign`, and especially if we only want to remove FAILED shards. It's certainly possible, even likely, that a removed assignment would be re-assigned to the same consumer process -- in which case we might inadvertently remove it twice.

The journal & shards Apply RPCs thread through an ExpectedModRevision which is provided by the client, and informed by an initial list RPC, for this reason. Not sure that makes sense here (the List RPC doesn't return the ModRevision of assignments), but UnassignRequest _could_ represent the desire to remove only FAILED shards, and this would need to guard against the possibility of raced invocations.

Let me know if you want to talk it out in person.
___
*[consumer/protocol/protocol.proto, line 450 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr06QHBDCgVwJIGiMbR:-Mr3Mn1w8jYyQuojLDsy:b6tg31b) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/protocol/protocol.proto#L450)):*
<details><summary><i>Previously, saterus (Alex Burkhart) wrote…</i></summary><blockquote>

Sure thing. It raises the question of error handling though. Would you expect the assignment removals to be transactional? Or if we encounter an error with one of many, should we just bail?
</blockquote></details>

Transactional. I'd approach it by building up a single Etcd transaction which composes invariant checks and desired deletions.

saterus (commented):
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312)* status: 2 of 11 files reviewed, 7 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[cmd/gazctl/gazctlcmd/shards_unassign.go, line 52 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr02Gzm6MFhe6Oq-3Iq:-Mr34zXV4q_ysuKAdqt3:bupf4h1) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/shards_unassign.go#L52)):*

supernit: perhaps log a line for each shard? That can also include more information, like its primary status, and its number of route members.
</blockquote></details>

I've taken a stab at this one. If we used the old `listShard` results, it'll have the previous state of the shard, which isn't useful. If we query immediately, the primary is empty because the allocator hasn't had a chance to run yet. Presumably, this still shows useful info about other replicas though.

On the other hand, we can sleep for a couple of seconds until the allocator has had a chance to run. Then the primary will be reassigned. This shortcuts needing to run `shards list` immediately after running `shards unassign`. Presumably I could do something fancy with watching for the assignment to take place and wait for that to happen, but I'm not sure if that's a good tactic or worth the trouble.

What do you think?
___
*[consumer/shard_api.go, line 197 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04lgE0HXP7gnS6qcL:-Mr3JrmhEg8KmqfiykTn:brdxwz5) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L197)):*

You'll need to RLock the KeySpace mutex here, to guard access to Resolver.state
</blockquote></details>

I had this originally, but I didn't end up accessing the state with the "delete _all_ the assignments" solution I'd come up with. Forgot to add it back when I narrowed it down to the primary. :grimacing: 
___
*[consumer/shard_api.go, line 204 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04o4UEmXfzz-lgpZy:-Mr3CnHjD6-I_jFVfNaT:bhal8av) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L204)):*

If the primary shard has status < FAILED, then this should further check that:
 * The number of assignments matches the desired replication.
 * All shard assignments are consistent.

These are sanity checks which help sure the required replication invariants of the shard.
If the primary is FAILED, we can of course be more willing to remove it.

See:
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/allocator#IsConsistentFn
 - https://pkg.go.dev/go.gazette.dev/core@v0.89.0/consumer#ShardIsConsistent

Any other potential sanity checks you're aware of?
</blockquote></details>

Good call. Can an inconsistent shard _reach_ a consistent state if the primary fails?
___
*[cmd/gazctl/gazctlcmd/gazctl.go, line 74 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr01Ssg4yTzUCr3Vr3o:-Mr31dFmAyLQEk60umyM:brm4s7h) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/gazctl.go#L74)):*

nit: I think this can be inlined into `cmdShardsUnassign` (or you could re-use `pruneConfig`?). Other structs in here were all extracted because they're used in multiple places (typically for a journals vs shards flavor of an operation).

for now, or later: What about a boolean Failed which further restricts the selection to primary shards which have a status of `FAILED`, and leaves others alone ?
</blockquote></details>

Yep, it can be. I put this one here since it looked like the other configs were defined here as some sort of easy reference. I'll inline it.

Heh, I had `--failed` and then removed it. I'll add it back.

saterus:
Thanks for the review. I believe I've addressed everything you spotted in the first round. I think it's ready for another look.

jgraettinger (approved):
LGTM % comments

> Reviewed 8 of 9 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312#-:-MrOWDH06hg71l8URKRZ:b-tz1f3e)* status: all files reviewed, 3 unresolved discussions (waiting on @saterus)

___
*[cmd/gazctl/gazctlcmd/shards_unassign.go, line 52 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr02Gzm6MFhe6Oq-3Iq:-MrOTD3UAHYoSI50976x:b-90hn7z) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/shards_unassign.go#L52)):*
<details><summary><i>Previously, saterus (Alex Burkhart) wrote…</i></summary><blockquote>

I've taken a stab at this one. If we used the old `listShard` results, it'll have the previous state of the shard, which isn't useful. If we query immediately, the primary is empty because the allocator hasn't had a chance to run yet. Presumably, this still shows useful info about other replicas though.

On the other hand, we can sleep for a couple of seconds until the allocator has had a chance to run. Then the primary will be reassigned. This shortcuts needing to run `shards list` immediately after running `shards unassign`. Presumably I could do something fancy with watching for the assignment to take place and wait for that to happen, but I'm not sure if that's a good tactic or worth the trouble.

What do you think?
</blockquote></details>

Neat! This seems useful, but the sleep does feel like a smell, and the subcommand is probably doing too much since the user could also do a follow-up `list` with the same selector to get this. The prior primary & status might be more interesting to log since that state is no longer query-able after the unassign. Again, we're in supernit territory here though.
___
*[consumer/shard_api.go, line 204 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr04o4UEmXfzz-lgpZy:-MrOUWvX16plhXmEDDS8:b-k7ilkt) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/consumer/shard_api.go#L204)):*
<details><summary><i>Previously, saterus (Alex Burkhart) wrote…</i></summary><blockquote>

Good call. Can an inconsistent shard _reach_ a consistent state if the primary fails?
</blockquote></details>

Yep: https://github.com/gazette/core/blob/v0.89.0/consumer/key_space.go#L68

A failed primary assignment is always consistent, and a failed replica is consistent if and only if the primary has also failed.
___
*[consumer/shard_api.go, line 207 at r3](https://reviewable.io/reviews/gazette/core/312#-MrOUu_E2j-9FXSqW2PC:-MrOUu_E2j-9FXSqW2PD:b-ul16zz) ([raw file](https://github.com/gazette/core/blob/a985cb5ec17ee433626ffa1af2d13a95bbec02bb/consumer/shard_api.go#L207)):*
> ```Go
> 			return resp, err
> 		} else if req.OnlyFailed && primaryAssignment.AssignmentValue.(*pc.ReplicaStatus).Code != pc.ReplicaStatus_FAILED {
> 			return resp, nil
> ```

this should be a continue right? We're not skipping _all_ shards, just this one?

saterus (commented):
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[cmd/gazctl/gazctlcmd/shards_unassign.go, line 52 at r1](https://reviewable.io/reviews/gazette/core/312#-Mr02Gzm6MFhe6Oq-3Iq:-MrOWqiw3JByO6MLJFGl:b7eioxq) ([raw file](https://github.com/gazette/core/blob/89ed8cc9094188e624aca39c7ed32af2f7ca06a8/cmd/gazctl/gazctlcmd/shards_unassign.go#L52)):*

Neat! This seems useful, but the sleep does feel like a smell, and the subcommand is probably doing too much since the user could also do a follow-up `list` with the same selector to get this. The prior primary & status might be more interesting to log since that state is no longer query-able after the unassign. Again, we're in supernit territory here though.
</blockquote></details>

I'll swap it out for the prior assignments. That's easy and doesn't require the sleep.
___
*[consumer/shard_api.go, line 207 at r3](https://reviewable.io/reviews/gazette/core/312#-MrOUu_E2j-9FXSqW2PC:-MrOX3ZY2UMk5N-1KLDC:b1rbozi) ([raw file](https://github.com/gazette/core/blob/a985cb5ec17ee433626ffa1af2d13a95bbec02bb/consumer/shard_api.go#L207)):*

this should be a continue right? We're not skipping _all_ shards, just this one?
</blockquote></details>

Yep, good catch. I'll add a test for this case.

saterus:
@jgraettinger I believe I've addressed the test failures and your latest concerns. I think this is ready for a final look.

jgraettinger (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gazette/core/312)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @saterus)

jgraettinger (approved):
Still LGTM

> *[Reviewable](https://reviewable.io/reviews/gazette/core/312#-:-MrSydxf6F5bHnD-DjZU:b-kwu2jz)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @saterus)

PR #313, opened 2022-01-06
Improve Unassign Shard Workflow

saterus:
When running `shards unassign --failed` against a cluster with many tasks, it previously would report all non-failed Shards as being successfully unassigned. This is flat wrong. It gracefully skipped those.

However, if there was a shard without an assignment, it would fail to work at all. Instead, exiting with an error. We should avoid that for the `--failed` case, but keep that behavior otherwise.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/313)


saterus:
We archived the Estuary fork and this PR cannot be updated. Replacing with a new PR.

PR #315, opened 2022-02-02
consumer: shard transitions account for delete-then-create race

### LINKED ISSUE #314: consumer: ABA concurrency between shard watchdog & transition handler
We (Estuary) run a shard watchdog job in our cluster which removes assignments for FAILED shards periodically.

Here's a scenario we recently observed:

* A shard failed, and is marked as FAILED
* Some time later, the shard-watchdog job noticed it was FAILED and removed the assignment.
* The consumer coordinator noticed the assignment removal and immediately re-assigned it back to the same pod where it had been running.
* The pod's Etcd keyspace watch loop (which has a deliberate Nagle-like delay of ~30ms) didn't notice that the assignment was removed and re-created. Specifically, there wasn't a keyspace watch loop iteration where it was able to notice the key didn't exist and take action accordingly, and by the time it did run, the key already existed again.

Solution: The resolver's [updateLocalShards](https://github.com/gazette/core/blob/master/consumer/resolver.go#L272-L310) function needs to account for the assignment's Etcd creation revision. If the prior and current creation revision differ, that's an implicit delete-then-create operation.

jgraettinger:
If a shard assignment is removed (for example, because its FAILED)
and is then immediately re-created by the allocation coordinator,
we previously could fail to notice that the prior local shard was
invalidated and should be re-created.

Account for this by additionally monitoring the CreateRevision of
assignments and treating an assignment of the same or higher slot,
with a differing CreateRevision, as being a novel assignment which
requires a novel local shard to be boot-strapped.

Fixes #314

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/315)

### DIFF
--- a/consumer/resolver.go
+++ b/consumer/resolver.go
@@ -279,10 +279,25 @@ func (r *Resolver) updateLocalShards() {
 	var next = make(map[pc.ShardID]*shard, len(r.state.LocalItems))
 
 	for _, li := range r.state.LocalItems {
-		var assignment = li.Assignments[li.Index]
-		var id = pc.ShardID(li.Item.Decoded.(allocator.Item).ID)
+		var (
+			asn       = li.Assignments[li.Index]
+			id        = pc.ShardID(li.Item.Decoded.(allocator.Item).ID)
+			shard, ok = r.shards[id]
+		)
 
-		var shard, ok = r.shards[id]
+		if !ok {
+			// Local shard doesn't exist and must be created.
+		} else if p := shard.resolved.assignment; asn.Decoded.(allocator.Assignment).Slot < p.Decoded.(allocator.Assignment).Slot {
+			// Local shard is being promoted (update).
+		} else if p.Raw.CreateRevision == asn.Raw.CreateRevision {
+			// Assignment value was updated but the assignment itself is unchanged.
+		} else {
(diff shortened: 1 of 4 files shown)


consumer/resolver.go
```diff
@@ -279,10 +279,25 @@ func (r *Resolver) updateLocalShards() {
...
+			id        = pc.ShardID(li.Item.Decoded.(allocator.Item).ID)
+			shard, ok = r.shards[id]
+		)
+
+		if !ok {
+			// Local shard doesn't exist and must be created.
+		} else if p := shard.resolved.assignment; asn.Decoded.(allocator.Assignment).Slot < p.Decoded.(allocator.Assignment).Slot {
+			// Local shard is being promoted (update).
+		} else if p.Raw.CreateRevision == asn.Raw.CreateRevision {
+			// Assignment value was updated but the assignment itself is unchanged.
```
saterus, line 293: I'm not sure I understand this case. How would the assignment value be updated without creating a new revision?

saterus (approved):

consumer/resolver.go
jgraettinger: [see the etcd spec](https://github.com/etcd-io/etcd/blob/main/api/mvccpb/kv.proto#L12-L29). CreateRevision is separately tracked from ModRevision.

PR #316, opened 2022-02-04
Improve Unassign Shard Workflow

saterus:
When running shards unassign --failed against a cluster with many tasks, it previously would report all non-failed Shards as being successfully unassigned. This is flat wrong. It gracefully skipped those.

However, if there was a shard without an assignment, it would fail to work at all. Instead, exiting with an error. We should avoid that for the --failed case, but keep that behavior otherwise.

This replaces #313, as it was connected to a branch on the Estuary fork.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/316)


jgraettinger (approved):
LGTM % comments

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gazette/core/316#-:-Mv5q9Pn0AY_ty1fHryA:b-tz1f3e)* status: all files reviewed, 2 unresolved discussions (waiting on @saterus)

___
*[cmd/gazctl/gazctlcmd/shards_unassign.go, line 61 at r1](https://reviewable.io/reviews/gazette/core/316#-Mv5nNI7BMLXeNentANB:-Mv5nNI89tdDHTWQb8JC:b-fee4kb) ([raw file](https://github.com/gazette/core/blob/a6e1180b491cab40c639f1707f8a8bca473c24ce/cmd/gazctl/gazctlcmd/shards_unassign.go#L61)):*
> ```Go
> 	} else {
> 		for _, shardId := range unassignResp.Shards {
> 			var shard = shardsById[shardId]
> ```

nit: this seems to be the only usage of `shardsById`. Why not just range over `listResp.Shards` and drop `shardsById` (and perhaps `shards`) altogether?
___
*[consumer/shard.go, line 166 at r1](https://reviewable.io/reviews/gazette/core/316#-Mv5nti9-JH4Bln0gRn6:-Mv5nti9-JH4Bln0gRn7:b-4p8o6k) ([raw file](https://github.com/gazette/core/blob/a6e1180b491cab40c639f1707f8a8bca473c24ce/consumer/shard.go#L166)):*
> ```Go
> }
> 
> func disableShardTransitions() (reset func()) {
> ```

This is test code and should be in `test_support_test.go`.

It's also a bit out of date now with https://github.com/gazette/core/commit/0087160ef15e0a9ca96b9b25745214eef1620c0b -- notably you don't need to assign `spec` and `assignment`.

There's also a (relatively new) `testing.T.Cleanup()` method which can be used to defer the restoration, which is probably a better practice than returning a closure -- a pattern which predates Cleanup.

PR #319, opened 2022-04-22
Add DeferPublishUncommitted to message.Publisher

psFried:
This is an experimental API to allow for publishing a message where the
content is not known until after the acknowledgement intents are needed.
This is a low level API to allow consumers a little more flexibility
when storing checkpoints in external systems, as Flow does for
materializations.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/319)

### DIFF
--- a/message/publisher.go
+++ b/message/publisher.go
@@ -128,6 +128,97 @@ func (p *Publisher) PublishUncommitted(mapping MappingFunc, msg Message) (*clien
 	return aa, nil
 }
 
+// PendingPublish is returned from DeferPublishUncommitted, and allows appending a single message
+// that had previously been sequenced.
+//
+// **This is a new and unstable API, that is subject to breaking changes.**
+type PendingPublish struct {
+	publisher   *Publisher
+	journal     pb.Journal
+	contentType string
+	uuid        UUID
+}
+
+// Resolve completes a PendingPublish by appending the finalized content of a message that had
+// previously been sequenced. See DeferPublishUncommitted docs for more.
+//
+// **This is a new and unstable API, that is subject to breaking changes.**
+func (pf *PendingPublish) Resolve(msg Message) error {
+	if pf.publisher == nil {
+		// Sanity check for if Resolve has already been called, or if PendingPublish is zero-valued
+		// due to SequenceFutureMessage having returned an error.
+		panic("Pending publish has already been resolved")
+	}
+	if v, ok := msg.(Validator); ok {
+		if err := v.Validate(); err != nil {
+			return err
+		}
+	}
+	msg.SetUUID(pf.uuid)
+
+	var framing, err = FramingByContentType(pf.contentType)
+	if err != nil {
+		return err
+	}
+
+	var aa = pf.publisher.ajc.StartAppend(pb.AppendRequest{Journal: pf.journal}, nil)
+	aa.Require(framing.Marshal(msg, aa.Writer()))
+	err = aa.Release()
+	pf.publisher = nil // so that we can sanity check that Resolve isn't called twice
+	return err
+}
+
+// DeferPublishUncommitted is used to sequence a message that will be published at some future
+// point, but before the end of the transaction. It returns a PendingPublish, which can be resolved
+// by passing it the actual message to be published. This is used in situations where you need to
+// transactionally publish a message when you don't have the content of that message until after the
+// ack intents are built. This is an advanced, low level api, and care must be taken to use it
+// correctly to avoid corruption of journal content.
+//
+// The journal and contentType must be known up front, and the acknowledgement Message must also be
+// provided by the caller. It's up to the caller to ensure that these things are correct and
+// consistent.
+//
+// The returned PendingPublish does not need to ever be resolved, and can be dropped with no harm
+// done. If Resolve is called, then it must be called _before_ the acknowledgements are written.
+// Otherwise the resolved message will be ignored by ReadCommitted consumers. Also note that the
+// PendingPublish is not safe to Resolve concurrently with other uses of a Publisher.
+//
+// No other messages should be published to the journal using PublishUncommitted or PublishCommitted
+// before the PendingPublish is resolved. It it permissible to publish more than one message using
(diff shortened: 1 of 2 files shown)


jgraettinger (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gazette/core/319)* status: all files reviewed, 4 unresolved discussions (waiting on @psFried)

___
*[message/publisher.go, line 188 at r1](https://reviewable.io/reviews/gazette/core/319#-N0H-Fvz0sAfnlBq_KlJ:-N0H-AI5A_i_mXL5pX5Z:bemzywt) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher.go#L188)):*
> ```Go
> //
> // **This is a new and unstable API, that is subject to breaking changes.**
> func (p *Publisher) DeferPublishUncommitted(journal pb.Journal, contentType string, ack Message) (fut PendingPublish, err error) {
> ```

We're already required to pass in a Message for its ack-intents behavior. Should we also use it to determine `journal` and `contentType` as the regular PublishUncommitted does ? It would be up to the caller to fill out "enough" of the Message that that routing can be done, even if the final content of the Message isn't ready, but that seems fine and in-keeping with intended usages?

It does have further implications on the API though: should PendingPublish then capture the Message to be published, such that Resolve() doesn't require it by-argument ?

Seems like this might be simpler overall, since the caller has to have meaningful prior knowledge of what the Message instance is going to be anyway in order for the ACK intents behavior to work properly. The pattern would essentially be to allocate it and fill out what's known know, then start a deferred publish, then fill out the rest, then resolve.
___
*[message/publisher.go, line 204 at r1](https://reviewable.io/reviews/gazette/core/319#-N0H-i_-9VNa__P9YQ1h:-N0H-i_-9VNa__P9YQ1i:b21nbu9) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher.go#L204)):*
> ```Go
> 		p.intents = append(p.intents, AckIntent{
> 			Journal: journal,
> 			msg:     ack,
> ```

I recognize this is explicitly documented to be an acknolwedgement, but think this should still probably do a`ack.NewAcknowledgement(...)` as happens in `PublishUncommitted`.
___
*[message/publisher_test.go, line 218 at r1](https://reviewable.io/reviews/gazette/core/319#-N0H0G3-Aoj6OeCNWCgM:-N0H0G3-Aoj6OeCNWCgN:by0dfmf) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher_test.go#L218)):*
> ```Go
> }
> 
> func TestDeferPublishUncommitted(t *testing.T) {
> ```

great test!
___
*[message/publisher_test.go, line 244 at r1](https://reviewable.io/reviews/gazette/core/319#-N0Gzc3u6vGvG2IUnQiu:-N0Gzc3u6vGvG2IUnQiv:bet9icx) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher_test.go#L244)):*
> ```Go
> 		require.NoError(t, err)
> 
> 		if seq.QueueUncommitted(env) == QueueAckCommitReplay {
> ```

nit: you can probably increase the ring buffer size and panic in this case unless you're deliberately wanting to exercise ring buffer semantics (which seem unrelated to the unit under test) ?

psFried:
> We're already required to pass in a Message for its ack-intents behavior. Should we also use it to determine journal and contentType as the regular PublishUncommitted does ?

The problem that I see with that is that `Message` doesn't specify any way to mutate it besides setting the uuid, which would make for a really awkward interface for `Resolve`. How would the caller mutate the message before it's written? Something like:

```go
func (p *PendingPublish) Resolve(finishMessage func(Message) Message) error ...
```

would mean that the caller has to use a type assertion to get their concrete message type, then mutate it and return it. It's weird, too, because `Message` is an interface so you wouldn't normally pass `*Message` to mutate it, but would need to have `finishMessage` accept it by value and then return a (possibly different) Message. I guess we _could_ pass `*Message`, but it feels like something that would be frowned upon. WDYT?

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/gazette/core/319)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[message/publisher.go, line 188 at r1](https://reviewable.io/reviews/gazette/core/319#-N0H-Fvz0sAfnlBq_KlJ:-N0I4Bn11fFdxLsBSFgJ:b-896fix) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher.go#L188)):*

We're already required to pass in a Message for its ack-intents behavior. Should we also use it to determine `journal` and `contentType` as the regular PublishUncommitted does ? It would be up to the caller to fill out "enough" of the Message that that routing can be done, even if the final content of the Message isn't ready, but that seems fine and in-keeping with intended usages?

It does have further implications on the API though: should PendingPublish then capture the Message to be published, such that Resolve() doesn't require it by-argument ?

Seems like this might be simpler overall, since the caller has to have meaningful prior knowledge of what the Message instance is going to be anyway in order for the ACK intents behavior to work properly. The pattern would essentially be to allocate it and fill out what's known know, then start a deferred publish, then fill out the rest, then resolve.
</blockquote></details>

Done.
___
*[message/publisher_test.go, line 244 at r1](https://reviewable.io/reviews/gazette/core/319#-N0Gzc3u6vGvG2IUnQiu:-N0I4QFC6LnENyQLBAAA:b-g20zrk) ([raw file](https://github.com/gazette/core/blob/85849128d98b637ffd613ec2e9a639a5f9bd7aac/message/publisher_test.go#L244)):*

nit: you can probably increase the ring buffer size and panic in this case unless you're deliberately wanting to exercise ring buffer semantics (which seem unrelated to the unit under test) ?
</blockquote></details>

good call, that was a bit of copypasta left over

PR #322, opened 2022-05-02
`shards unassign` command no longer stalls

saterus:
This fixes a bug which caused the shard supervisor to stall indefinitely. If it ran `shards unassign` and there was nothing to be unassigned, it would never increment the etcd revision with its transaction. Then when we waited to observe our changes as reflected by the etcd mod revision, we would end up waiting until the next arbitrary transaction to complete. Needless to say, this is not the desired behavior.

Also adds a timeout to the command more generally.

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/322)

### DIFF
--- a/consumer/shard_api.go
+++ b/consumer/shard_api.go
@@ -234,7 +234,9 @@ func ShardUnassign(ctx context.Context, srv *Service, req *pc.UnassignRequest) (
 	etcdResp, err := srv.Etcd.Txn(ctx).If(cmp...).Then(ops...).Commit()
 	if err != nil {
 		return resp, fmt.Errorf("executing etcd transaction: %w", err)
-	} else if etcdResp.Succeeded {
+	} else if !etcdResp.Succeeded {
+		resp.Status = pc.Status_ETCD_TRANSACTION_FAILED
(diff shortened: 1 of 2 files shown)


consumer/shard_api.go
```diff
@@ -128,6 +128,11 @@ func ShardApply(ctx context.Context, srv *Service, req *pc.ApplyRequest) (*pc.Ap
 		}
 	}
 
+	if len(ops) == 0 {
```
jgraettinger, line 131: I'm missing why this would be required given the check of `len(ops)` below. The current behavior is to go ahead and do a no-op etcd transaction, but to only wait on the revision if there was at least one applied operation.

I think you meant this to be part of `ShardUnassign`, below? For consistency I'd recommend having the same handling in `ShardUnasign` as is currently done with `ShardApply` (do the no-op transaction, but only wait if empty).

saterus: Doh. I don't know how I ended up modifying `ShardApply`, but that wasn't the intention. :facepalm:

jgraettinger (approved):
Yes, apologies. LGTM

PR #323, opened 2022-05-13
parse SSEKMSKeyId from store url and pass to store config

davidgmonical:
Adds in support for `SSEKMSKeyId` by parsing it from the store url and passing it into the `S3StoreConfig`. Done analogously to the `SSE` paramater.

Also added a unit test for `parseStoreArgs` which checks that the `SSEKMSKeyId` field (amongst others) is correctly parsed from the store URL

Additional Testing:

- Ran `TestStoreInteractions` using `SSE` and `SSEKMSKeyId` parameters on a bucket with SSE KMS + key. Used this structure as the fragment store:  `s3://bucket-with-kms-key/?find=rwFind&replace=rwReplace&endpoint=https://region.amazonaws.com&sse=aws:kms&ssekmskeyid=test-key")` where  `test-key` matched the one on `bucket-with-kms-key`. The test ran successfully and was able to `Persist, List, Open/Read, and Delete` objects from the bucket. 

- With the above setup, I removed the deletion portion of the test so that the encrypted objects would remain in the bucket. I then attempted through the CLI to call `get-objects`, once with a role without permissions to decrypt the objects and once with a role with permission to decrypt the objects. (*Note the kms key does not actually need to be passed in, rather the role calling the endpoint needs to have permissions to decrypt the object ([source](https://aws.amazon.com/premiumsupport/knowledge-center/decrypt-kms-encrypted-objects-s3/))). 
The result for the role without decryption perms was:
`An error occurred (AccessDenied) when calling the GetObject operation: The ciphertext refers to a customer master key that does not exist, does not exist in this region, or you are not allowed to access. ` 
And the result for the role with decryption perms was a successful download.
This is in evidence of the `Persist` correctly using the `SSEKMSKeyId` passed in from the fragment store to authorize decryption

 

---
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/323)

### DIFF
--- a/broker/fragment/store_s3.go
+++ b/broker/fragment/store_s3.go
@@ -41,6 +41,9 @@ type S3StoreConfig struct {
 	// SSE is the server-side encryption type to be applied (eg, "AES256").
 	// By default, encryption is not used.
 	SSE string
+	// SSEKMSKeyId specifies the ID for the AWS KMS symmetric customer managed key
+	// By default, not used.
+	SSEKMSKeyId string
 }
 
 type s3Backend struct {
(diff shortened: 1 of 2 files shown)


psFried (approved):
Thanks for the PR! I think this looks good to merge.

PR #327, opened 2022-08-25
Allow customizing server Listener

psFried:
Adds a function to server.go that allows passing the `net.Listener` used
with starting a Server. This is intended to allow users to create
servers using TLS.

Note that this is only a library change. No TLS options were added to Gazette binaries as of yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/327)


jgraettinger (approved):
LGTM

PR #329, opened 2023-01-06
consumer: synchronously fail if StartCommit returns a pre-resolved error

jgraettinger:
This ensures txnRun fails-fast if StartCommit returns an error. If we don't do this, then txnStep will race a select over the shard's `readCh` with a select of the (failed) OpFuture, and then go on to call ConsumeMessage of a shard which is already in a failing state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/329)

### DIFF
--- a/consumer/transaction.go
+++ b/consumer/transaction.go
@@ -443,7 +443,21 @@ func txnStartCommit(s *shard, txn *transaction) (pc.Checkpoint, error) {
 	var waitFor = s.ajc.PendingExcept(s.recovery.log)
 
 	trace.Log(s.ctx, "store.StartCommit", s.resolved.fqn)
-	txn.commitBarrier = s.store.StartCommit(s, txn.checkpoint, waitFor)
+	var barrier = s.store.StartCommit(s, txn.checkpoint, waitFor)
+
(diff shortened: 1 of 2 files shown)


consumer/transaction.go
```diff
@@ -443,7 +443,21 @@ func txnStartCommit(s *shard, txn *transaction) (pc.Checkpoint, error) {
...
+	// failed.
+	select {
+	case <-barrier.Done():
+		if barrier.Err() != nil {
+			return pc.Checkpoint{}, fmt.Errorf("store.StartCommit: %w", barrier.Err())
+		}
+	default:
+	}
+
+	txn.commitBarrier = barrier
```
psFried, line 460: The `txn.commitBarrier` now won't be set if we're synchronously returning the error. Are all the other accesses of `commitBarrier` OK with that? Should it clear the previous value? I looked through the file and didn't see any problems, but just wanted to double check.

psFried (approved):
LGTM

consumer/transaction.go
jgraettinger: yep 👍

PR #336, opened 2023-04-12
Fix keyspace missing watch events

### LINKED ISSUE #337: Etcd watch events are sometimes missed
We've noticed a few instances where Gazette brokers have ended up with inconsistent views of journals. This coincided with a re-deployment of both our brokers and consumers. At this point, we seem to reproduce it in ~1 out of every 5-10 times we deploy. This also coincided with the following log line, which was the biggest clue:

```
2023-04-07T14:44:24.004413866Z stderr F {"cancelled":false,"compactRevision":0,"created":false,"header":{"cluster_id":9261053468663688537,"member_id":9752624356811177708,"revision":15595063,"raft_term":73},"isProgressNotify":false,"level":"warning","msg":"received duplicate Etcd watch revision (ignoring)","numEvents":2039,"time":"2023-04-07T14:44:24Z","watchIter":1}
```

This results in one or more brokers that refuse to exit gracefully.

psFried:
Fixes #337

It turns out that our assumptions about etcd watch responses weren't quite right. We had assumed that the `Header.Revision` would be the same as the highest `ModRevision` of the events within the response. While this happens to normally be the case, it is not guaranteed, and the revisions can differ significantly when the system is under heavy load. So we'd process a watch response that has a `Header.Revision` of `n`, but the most recent `ModRevision` of the contained `Events` was less than `n`. Then there's a subsequent watch response that _also_ has a `Header.Revision` of `n` that contains the remainder of the events up through revision `n`. But that second response gets ignored because the keyspace thinks it's already processed up through revision `n`.

This commit changes the way we process watch responses so that we only update the high water mark based on the `ModRevision` of the events, rather than just looking at the header. As a special case, we still use the `Header.Revision` when the response represents a progress notification, which has no inner events.

This also changes the keyspace to allow it to process the same revision multiple times. This was needed because when we re-start a watch, we now start at the last revision that's known to have been processed. The previous behavior was to always add 1 to the last known revision. So if events for a large transaction (which would all have the same `ModRevision`) get split over multiple watch responses, we should process them just fine.

I tested this locally by running the crash tests with the stream-sum example. I also tested by manually creating ~1k journals and then re-starting the brokers to trigger a bunch of watch events.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/336)

### DIFF
--- a/keyspace/key_space.go
+++ b/keyspace/key_space.go
@@ -23,11 +23,13 @@ import (
 type KeySpace struct {
 	// Key prefix which roots this KeySpace.
 	Root string
-	// Header is the last Etcd operation header which updated this KeySpace.
-	// It's important (for reasons I don't fully understand) that this revision is
-	// not updated by events that don't affect the range of keys in this keyspace.
-	// In other words, the `Header.Revision` should always be the greatest `ModRevision`
-	// of any key prefixed by the `Root`.
+	// Header is the last Etcd operation revision which actually updated this KeySpace:
+	// in other words, `Header.Revision` is the greatest `ModRevision` of any key under
+	// the `Root`. It is not incremented by other Etcd operations which don't actually
+	// change the KeySpace, such as progress notifications. This property means, for
+	// all watched KeySpaces having a common Root, their revisions are always directly
+	// comparable and will be equal if (and only if) they reflect the same keys and
+	// values.
 	Header etcdserverpb.ResponseHeader
 	// KeyValues is a complete and decoded mirror of the (prefixed) Etcd key/value space.
 	KeyValues
@@ -140,7 +142,7 @@ func (ks *KeySpace) Watch(ctx context.Context, client clientv3.Watcher) error {
 	// This is separate from the `ks.Header.Revision`, which tracks the latest
 	// revision that's actually modified the keyspace. ProgressNofify responses
 	// should update the `resumeRevision`, but _not_ the `ks.Header.Revision`.
-	var resumeRevision = ks.Header.Revision
+	var resumeRevision = ks.Header.Revision + 1
 	ks.Mu.RUnlock()
 	for attempt := 0; true; attempt++ {
 		// Start (or restart) a new long-lived Watch. Note this is very similar to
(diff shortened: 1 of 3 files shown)


keyspace/key_space.go
```diff
@@ -184,24 +190,32 @@ func (ks *KeySpace) Watch(ctx context.Context, client clientv3.Watcher) error {
...
-					"header":           resp.Header,
-					"isProgressNotify": resp.IsProgressNotify(),
-					"cancelled":        resp.Canceled,
-					"created":          resp.Created,
-					"compactRevision":  resp.CompactRevision,
-					"numEvents":        len(resp.Events),
-					"watchIter":        attempt,
-				}).Warn("received duplicate Etcd watch revision (ignoring)")
-			} else {
+					"header":          resp.Header,
```
jgraettinger, line 198: optional supernit:

this is starting to get into "preference", but it's grounded enough to share, so I'm doing so and do with it what you will:

In your shoes, I would inline the lookups that yield `minModRevision` and `maxModRevision` in the log statement.
Then, below, I would have resumeRevision separately be updated with a comment explaining why we're setting it to the maximum mod revision. (I'd also comment in the `else` block below, as to why we're using `resp.Header.Revision`.

This does a couple of things. First, it makes the log statement side-effect-less and easy to move or delete, with no dependence on other stack variables. Second, it keeps separation between SLOC which are "about debug logging" from lines that are actual algorithm, which I find easier as a reader. Right now, it's not immediately obvious what role `minRevision` and `resumeRevision` are playing in the algorithm. It's not clear that `minRevision` is _only_ for logging out, and conversely the update of `resumeRevision` has _great_ significance.

keyspace/key_space.go
```diff
@@ -255,20 +269,36 @@ func (ks *KeySpace) WaitForRevision(ctx context.Context, revision int64) error {
...
-		// KeySpace, and we track only modifying revisions.
-		if !wr.IsProgressNotify() {
-			if err := patchHeader(&hdr, wr.Header, false); err != nil {
-				return err
-			}
+		// Sanity check that the response header is consistent, without updating lastHeader.
+		if err := checkHeader(&lastHeader, wr.Header, true); err != nil {
+			return err
+		}
+		if len(wr.Events) == 0 {
```
jgraettinger, line 285: nit: this cannot ever happen. Just let it panic below instead if it were to.

keyspace/key_space.go
```diff
@@ -334,6 +364,14 @@ func (ks *KeySpace) onUpdate() {
 // expected to be greater than or equal to the current one; otherwise, it
 // should be strictly greater.
 func patchHeader(h *etcdserverpb.ResponseHeader, update etcdserverpb.ResponseHeader, allowSameRevision bool) error {
+	if err := checkHeader(h, update, allowSameRevision); err != nil {
```
jgraettinger, line 367: nit: `patchHeader` has just one remaining usage. I think the indirection is no longer serving us and it should be inlined to the remaining call site.

keyspace/key_space.go
```diff
@@ -334,6 +364,14 @@ func (ks *KeySpace) onUpdate() {
...
+		return err
+	}
+	*h = update
+	return nil
+}
+
+func checkHeader(h *etcdserverpb.ResponseHeader, update etcdserverpb.ResponseHeader, allowSameRevision bool) error {
```
jgraettinger, line 374: nit: `allowSameRevision` is now always true, and should be removed.

jgraettinger (commented):
Some nits below. I'm being _very_ nit-picky but this code is also foundational corner-stone of how distributed consensus works.

I've had a lingering concern about the implications of ModRevisions being spread across multiple WatchEvents. The biggest consideration is that means that a broker could advertise / run consensus over a KeySpace that only has _part_ of a revision applied, which would Be Very Bad. It would take very bad luck for this to happen, but it seemingly could, and is not something we can prevent because _we don't know_ if a future WatchEvent is coming with more inner events of the ModRevision.

So, reading docs on this, [I spotted this](https://etcd.io/docs/v3.2/learning/api/#watch-streams):

> Atomic - a list of events is guaranteed to encompass complete revisions; updates in the same revision over multiple keys will not be split over several lists of events.

This indicates, to me, that inner events of the same ModRevision will _never_ straddle across two separate WatchEvents, which is great. How should this code be updated to reflect and enforce this presumption? I'd like to be able to guarantee either a) complete atomicity, or b) that we crash.

psFried:
> This indicates, to me, that inner events of the same ModRevision will never straddle across two separate WatchEvents, which is great. How should this code be updated to reflect and enforce this presumption? I'd like to be able to guarantee either a) complete atomicity, or b) that we crash.

Yes, I believe that's correct. Events for a given ModRevision will always be sent in a single WatchResponse unless you explicitly ask for them to be split by setting the [`fragment` option](https://pkg.go.dev/go.etcd.io/etcd/client/v3#WithFragment) on the WatchRequest (which defaults to `false`). I added a commit to add explicit assertions that the ModRevision increases with each response. I addressed the rest of the feedback, too. LMK what you think.

jgraettinger:
Thanks, I'll take a review pass shortly. One comment I'm jotting down is that, currently, it would be **incorrect** to set the `fragment` option. That's because we need to guarantee that a revision is processed in it's _entirety_ before we can allow readers to observe it.

The client-exposed [WatchResponse](https://pkg.go.dev/go.etcd.io/etcd/client/v3#WatchResponse) doesn't provide a means to understand whether the current event contains a partial revision and there's another WatchResponse coming that we must wait for.

Which is a bit frustrating, because the protobuf [WatchResponse](https://pkg.go.dev/go.etcd.io/etcd/api/v3/etcdserverpb#WatchResponse) _does_ have a `Fragment` field which could tell us we need to wait for a subsequent event. I further confirmed it does the "right" thing of setting Fragment = false on [the very last event](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L582).

So, _if_ that were plumbed through, then KeySpace could be made correct for arbitrary fragmentation by simply awaiting a next WatchResponse if the current one's `Fragment` is set.

jgraettinger (approved):
I had some additional feedback, but just pushed up another commit rather than going round again. PTAL, and if it passes your soak setup, it LGTM.

PR #338, opened 2023-05-08
keyspace: update MemberId/RaftTerm on watch events

psFried:
When processing watch events, update the persisted MemberId and RaftTerm. The `checkHeader` function logs any changes to these fields, as it's useful information for debugging. The last round of keyspace changes introduced a bug where we were no longer updating those fields if they changed after the initial keyspace load. This resulted in logging a warning on every subsequent watch event. This commit fixes that so that we once again only log on the first observation of the change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/338)


jgraettinger (approved):
LGTM

PR #339, opened 2023-06-13
consumer: add Shard.PrimaryLoop() client.OpFuture

jgraettinger:
Expose the status of the shard's primary processing loop to those who might care.

This allows applications to reliably detect when a shard has failed, and is a more reliable mechanism than the BeginFinisher interface because that interface is only called in the context of a current consumer transaction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/339)


williamhbaker (approved):
LGTM

PR #341, opened 2023-06-28
feature: Implement `azure-ad://` as a fragment store URI scheme

jshearer:
This updates the Azure fragment store to support active directory authentication. In order to do so, you have to create an "Application" in Azure, and then generate a secret for that application (also called a service principal). In order to share access to storage accounts in external tenants, you must add the application to that tenant, and then grant it access to the storage account.

Three new environment variables are exposed here:
* `AZURE_TENANT_ID`: the name of your tenant
* `AZURE_CLIENT_ID`: the id of your application
* `AZURE_CLIENT_SECRET`: the secret you generated for your application

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/341)


broker/fragment/store_azure.go
```diff
@@ -29,30 +35,48 @@ type azureBackend struct {
...
+
+	// Set current and past time and create key
+	currentTime := time.Now().UTC().Add(-10 * time.Second)
+	expTime := currentTime.Add(d)
+	info := service.KeyInfo{
+		Start:  to.Ptr(currentTime.UTC().Format(sas.TimeFormat)),
+		Expiry: to.Ptr(expTime.UTC().Format(sas.TimeFormat)),
+	}
+
+	udc, err := a.svcClient.GetUserDelegationCredential(context.Background(), info, nil)
```
jgraettinger, line 62: This is a network request, right? If so, we'd need to root this out and reuse, as we really don't want that on the path here. This method should be _fast_.

It seems like the start / expiry of the delegation key, vs the start / expiry of the individual signatures, are being conflated. We can get a much longer lived delegation key and reuse it over and over.

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,48 @@ type azureBackend struct {
...
 		return "", err
 	}
 	blobName := cfg.rewritePath(cfg.prefix, fragment.ContentPath())
-	sasQueryParams, err := azblob.BlobSASSignatureValues{
-		Protocol:      azblob.SASProtocolHTTPS, // Users MUST use HTTPS (not HTTP)
-		ExpiryTime:    time.Now().UTC().Add(d), // Timestamps are expected in UTC https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas#service-sas-example
+
+	// Set current and past time and create key
+	currentTime := time.Now().UTC().Add(-10 * time.Second)
+	expTime := currentTime.Add(d)
```
jgraettinger, line 56: this isn't accounting for the -10 seconds taken from currentTime. Not that values of `d` can be pretty small here.

Is is _absolutely required_ that we set a `StartTime` below? We didn't used to.

broker/fragment/store_azure.go
```diff
@@ -167,6 +191,24 @@ func (a *azureBackend) Remove(ctx context.Context, fragment pb.Fragment) error {
...
+func GetAzureStorageCredential(coreCredential azcore.TokenCredential) (azblob.TokenCredential, error) {
+	var tokenRefresher azblob.TokenRefresher
+	tokenRefresher = func(credential azblob.TokenCredential) time.Duration {
+		accessToken, err := coreCredential.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{"https://storage.azure.com/.default"}})
+		if err != nil {
+			panic(err)
+		}
+		credential.SetToken(accessToken.Token)
+
+		exp := accessToken.ExpiresOn.Sub(time.Now().Add(2 * time.Minute))
```
jgraettinger, line 204: where is two minutes coming from?

broker/fragment/store_azure.go
```diff
@@ -183,17 +225,48 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
 	accountName := os.Getenv("AZURE_ACCOUNT_NAME")
-	accountKey := os.Getenv("AZURE_ACCOUNT_KEY")
-	credentials, err := azblob.NewSharedKeyCredential(accountName, accountKey)
-	if err != nil {
-		return
+	a.endpoint = fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+
+	if ep.Scheme == "azure" {
+		accountKey := os.Getenv("AZURE_ACCOUNT_KEY")
+		sharedKeyCred, cred_err := service.NewSharedKeyCredential(accountName, accountKey)
```
jgraettinger, line 235: nit: `cred_err` => `credErr` (Go is camel case). Or just `err`, there's no other meaningful error variable.

broker/fragment/store_azure.go
```diff
@@ -167,6 +191,24 @@ func (a *azureBackend) Remove(ctx context.Context, fragment pb.Fragment) error {
 	return err
 }
 
+// https://gist.github.com/ItalyPaleAle/ec6498bfa81a96f9ca27a2da6f60a770
```
jgraettinger, line 194: what is the purpose of this? it appears unused right now?

broker/fragment/store_azure.go
```diff
@@ -183,17 +225,48 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
-		return
+	a.endpoint = fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+
+	if ep.Scheme == "azure" {
+		accountKey := os.Getenv("AZURE_ACCOUNT_KEY")
+		sharedKeyCred, cred_err := service.NewSharedKeyCredential(accountName, accountKey)
+		if cred_err != nil {
+			return cfg, nil, cred_err
+		}
+		serviceClient, cred_err := service.NewClientWithSharedKeyCredential(a.endpoint, sharedKeyCred, &service.ClientOptions{})
```
jgraettinger, line 239: you're building this so that you can then make a rest request to get a delegation key? kind of circuitous in this case.
Why not just hang onto the credentials and keep the signing code path the same for `azure://` ? As a reader, I'd definitely prefer clearly delineated paths for the azure vs azure-ad cases.

It may also make sense to fetch a delegation key only once, below, when initializing for `azure-ad` and thereafter use it for signing credentials.

broker/fragment/store_azure.go
```diff
@@ -183,17 +225,48 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
+
 	accountName := os.Getenv("AZURE_ACCOUNT_NAME")
-	accountKey := os.Getenv("AZURE_ACCOUNT_KEY")
-	credentials, err := azblob.NewSharedKeyCredential(accountName, accountKey)
-	if err != nil {
-		return
+	a.endpoint = fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+
+	if ep.Scheme == "azure" {
+		accountKey := os.Getenv("AZURE_ACCOUNT_KEY")
```
jgraettinger, line 234: nit: code style in this repo is to prefer `var accountKey = ...` over `accountKey := ` for new variable initialization.

jgraettinger (commented):
thanks, some comments below

broker/fragment/store_azure.go
jshearer: Yuuup, good point. Updated to fetch this when it's expiring soon, otherwise return the cached value

broker/fragment/store_azure.go
jshearer: Probably not. I'll test with it removed

broker/fragment/store_azure.go
jshearer: This is a padding window to make sure we refresh the access token before it expires. 2 minutes seems reasonable, [given that](https://learn.microsoft.com/en-us/azure/active-directory/develop/configurable-token-lifetimes#access-tokens): 

> The default lifetime of an access token is variable. When issued, an access token's default lifetime is assigned a random value ranging between 60-90 minutes (75 minutes on average)

broker/fragment/store_azure.go
jshearer: It's a link to a gist where I learned about the TokenRefresher concept, but it's not super important. Removed

broker/fragment/store_azure.go
jshearer: Removed

broker/fragment/store_azure.go
jshearer: Not entirely sure what you're saying here. We need something that implements `service.Client`, since that's how we get a `UserDelegationCredential`, which we need to build signed URLs. When we're using shared keys (`azure://`), we  [call `NewClientWithSharedKeyCredential()`](https://github.com/gazette/core/pull/341/files#diff-b58442da8fead03ddc65f71858da8439f9f4b987fc32da9b453d9fba6a066c88R252):, and when we're using service principal auth (`azure-ad://`), we [use `NewClient()`](https://github.com/gazette/core/pull/341/files#diff-b58442da8fead03ddc65f71858da8439f9f4b987fc32da9b453d9fba6a066c88R273) and give it the custom token refresher built by `GetAzureStorageCredential()`.

The idea being in the end, we have something that looks like a `service.Client`, which we can call `GetUserDelegationCredential` on, no matter whether we're using `azure://` or `azure-ad://`.

> It may also make sense to fetch a delegation key only once, below, when initializing for azure-ad and thereafter use it for signing credentials.

Yup, I refactored this out into a function that generates the delegation key when it's about to expire, otherwise returns a cached one.

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,63 @@ type azureBackend struct {
...
+	udc         *service.UserDelegationCredential
+	udc_exp     *time.Time
 }
 
 func (a *azureBackend) Provider() string {
 	return "azure"
 }
 
+// Cache UserDelegationCredentials and refresh them when needed
+func (a *azureBackend) GetUserDelegationCredential() (*service.UserDelegationCredential, error) {
```
jgraettinger, line 49: this should be lower-case so it's not exported. Also, it should live near the bottom as utility / implementation detail. Important code first.

Also, does this need to have an expiration time?  Is that enforced by Azure ? Are these minting unique credential keys? Could we mint one that has the process's lifetime? We never write these out or transmit them anywhere so I'm unclear what the short expiration is seeking to protect but perhaps Azure requires it?

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,63 @@ type azureBackend struct {
 	endpoint    string
 	accountName string
 	client      pipeline.Pipeline
-	credentials *azblob.SharedKeyCredential
+	svcClient   service.Client
 	clientMu    sync.Mutex
+	udc         *service.UserDelegationCredential
+	udc_exp     *time.Time
```
jgraettinger, line 41: udc_exp should be camel case.

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,63 @@ type azureBackend struct {
...
 }
 
 func (a *azureBackend) Provider() string {
 	return "azure"
 }
 
+// Cache UserDelegationCredentials and refresh them when needed
+func (a *azureBackend) GetUserDelegationCredential() (*service.UserDelegationCredential, error) {
+	if a.udc == nil || (a.udc_exp != nil && a.udc_exp.After(time.Now().Add(time.Minute*10))) {
+		expTime := time.Now().Add(time.Hour * 24)
```
jgraettinger, line 51: nit: please see the other comment I wrote about code style in this repo being `var variableName =`

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,63 @@ type azureBackend struct {
...
 	cfg, _, err := a.azureClient(ep)
 	if err != nil {
 		return "", err
 	}
 	blobName := cfg.rewritePath(cfg.prefix, fragment.ContentPath())
-	sasQueryParams, err := azblob.BlobSASSignatureValues{
-		Protocol:      azblob.SASProtocolHTTPS, // Users MUST use HTTPS (not HTTP)
+
+	udc, err := a.GetUserDelegationCredential()
+
```
jgraettinger, line 79: nit: extra newline

broker/fragment/store_azure.go
```diff
@@ -167,6 +206,22 @@ func (a *azureBackend) Remove(ctx context.Context, fragment pb.Fragment) error {
 	return err
 }
 
+func GetAzureStorageCredential(coreCredential azcore.TokenCredential) (azblob.TokenCredential, error) {
```
jgraettinger, line 209: My comment you resolved was in reference to this entire routine, not the URL comment.

It's not used anywhere that I can tell. If you do intend for it to be exported, it needs a documenting comment for what it is, what it does, why it exist.

broker/fragment/store_azure.go
jgraettinger: Okay, that needs a comment.

broker/fragment/store_azure.go
```diff
@@ -183,17 +238,48 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
+			return cfg, nil, err
+		}
+		a.svcClient = *serviceClient
+		// azblob SharedKeyCredential is apparently different from service SharedKeyCredential
+		credentials, err = azblob.NewSharedKeyCredential(accountName, accountKey)
+		if err != nil {
+			return cfg, nil, err
+		}
+	} else if ep.Scheme == "azure-ad" {
+		// https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication-service-principal?tabs=azure-cli#-option-1-authenticate-with-a-secret
```
jgraettinger, line 263: nit: See other comment about free-floating URLs like this.

broker/fragment/store_azure.go
```diff
@@ -29,30 +35,63 @@ type azureBackend struct {
...
+		}
+
+		a.udc = udc
+		a.udc_exp = &expTime
+	}
+
+	return a.udc, nil
+}
+
+// https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/service/examples_test.go#L285
```
jgraettinger, line 70: nit: when dropping URLs like this, include _why_ the URL is there. Give me, as a reader, some sense of what I might find there.

jgraettinger (commented):
Some more comments.

As a meta-comment: please do a bit more self-review for code style / consistency / formatting / comments etc.

broker/fragment/store_azure.go
jgraettinger: > ... since that's how we get a UserDelegationCredential, which we need to build signed URLs

But we don't need a `UserDelegationCredential` in the `azure://` case, that's what I was getting at. That's needed only for the azure-ad case. Before this PR, we would build the structure for signing, and then simply called `.NewSASQueryParameters(a.credentials)`.

The question i'm asking is: does use of a `UserDelegationCredential` represent an improvement for the `azure://` case? It's downside is that it's an obfuscation of what's actually happening in the signing path, and thus _could_ be reserved for use only in the `azure-ad://` case.

I do see that the structures for signing look _nearly_ identical but are technically different structs in different packages. I'm unclear what that's about, or if one structure is actually implemented in terms of the other or something, and whether that's something we could leverage.

I'm not coming down hard on one way vs the other, I'm just looking for "We considered doing ABC but went with XYZ because ..."

broker/fragment/store_azure.go
jshearer: It needs an expiry. This is creating a signing key that lets you statelessly sign specific blob requests, so I assume they want expiration for the same reason e.g a JWT needs expiration: the fallout of a time-limited key getting leaked is less than an unlimited key.

[The docs](https://learn.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key#request-body) say:

Element | Description
-- | --
Start | Required. The start time for the user delegation SAS, in ISO date format. It must be a valid date and time within seven days of the current date.
Expiry | Required. The expiration time of the user delegation SAS, in ISO date format. It must be a valid date and time within seven days of the current date.

jshearer: Changed to lower case and moved near the bottom

broker/fragment/store_azure.go
jshearer: Ok sorry... is that for all variables? I'm still not super familiar with these patterns in Go, but I'm seeing a bunch of different variable declaration styles in this file from before I made any changes and was just following along with what I saw.

This one seems to be using `:=`: 
https://github.com/gazette/core/blob/fc3ba772a67798314b8fe54f1e4d2976e4c527f0/broker/fragment/store_azure.go#L41

This one is using `var` but not actually assigning a value until later:
https://github.com/gazette/core/blob/fc3ba772a67798314b8fe54f1e4d2976e4c527f0/broker/fragment/store_azure.go#L110-L118

This one taught me something new: apparently return values can be named, and those are treated as implicit declarations:
https://github.com/gazette/core/blob/fc3ba772a67798314b8fe54f1e4d2976e4c527f0/broker/fragment/store_azure.go#L170

Would you like me to change everything here to using `var`? Or is there a linter I'm missing or something?

broker/fragment/store_azure.go
jshearer: Normally I'd just make this change since it's a nit-pick, but in light of your meta-comment above about the burden of common understanding... my understanding was that Go has a blessed formatter that is supposed to take care of all of these stylistic "tabs vs spaces" style issues, and I ran it (and it moved some stuff around and updated some formatting, yay!).

Since it didn't catch this (or the variable thing above), I assumed that it was good to go. Since that's not the case, can you point me to some resource that lays out these things, so that I can learn where I'm supposed to put newlines or not? There are clearly newlines between other bits of code, so I'm guessing the heuristic here is "no newlines between an error-producing statement and the error-checking block" which seems perfectly reasonable, just not something I would have known.

broker/fragment/store_azure.go
jshearer: Yup, this one I missed as its functionality got replaced by `azidentity.NewClientSecretCredential`. I didn't get a warning about an unused function since it had a capitalized name, which TIL means an exported function. Removed!

jshearer:
My intent with the URL comments is that Azure's SDKs and APIs are pretty convoluted, and it took me a while to find the right places in the docs that describe what I need. I generally try to not write tautological comments, and I guess in this case a comment like `// Link to the Azure docs talking about AD auth` felt like it wasn't actually useful, since it's clearly a link to the azure docks for authentication, but I'm happy to adjust upwards my tolerance for comments if that's the pattern here.

broker/fragment/store_azure.go
jshearer: Added a comment

broker/fragment/store_azure.go
jshearer: Added a comment

broker/fragment/store_azure.go
jshearer: this particular line is moot because it's in that function that I should have removed, but I will add a comment over [here](https://github.com/gazette/core/pull/341/files#diff-b58442da8fead03ddc65f71858da8439f9f4b987fc32da9b453d9fba6a066c88R263) to the same effect (expiration buffer)

broker/fragment/store_azure.go
jshearer: > But we don't need a `UserDelegationCredential` in the azure:// case

It's somewhat opaque, but my understanding is that this is a way to create SASs that abstracts over whether your underlying credential is a Shared Key (`azure://`), or a Client Secret (`azure-ad://`). In both scenarios we end up with a `azblob/service.Client` which exposes `GetUserDelegationCredential()`. 

I'm not seeing how it creates any more obfuscation? I was actually pretty happy with how it turned out, as the only logic conditional on whether we have `azure://` or `azure://` is in `azureClient()`, at which point we end up with a `service.Client` either way and no longer have to worry about _how_ we're authenticated, just that we are.

psFried:
> I generally try to not write tautological comments, and I guess in this case a comment like // Link to the Azure docs talking about AD auth felt like it wasn't actually useful, since it's clearly a link to the azure docs for authentication, but I'm happy to adjust upwards my tolerance for comments like that if that's the pattern here.

I'll just chime in and say that links, even for big public companies like MS, frequently break. IMO it's best to provide a brief summary of the relevant info from the link, and just treat the link more like a citation where readers can go for more details. This also saves readers from needing to visit the link in order to know whether it's relevant to their current focus.

jshearer:
I may be running into a stumbling block here: it looks like the libraries we're using to do the meat of the actual fragment store operations on Azure blobs here are both no longer maintained and are segfaulting when trying to run this e2e locally..

* `azure-storage-blob-go`:
  > If you would like to access our latest Go SDK, please refer to the new preview azblob package [here](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob#readme).
* `azure-pipeline-go`: https://github.com/Azure/azure-pipeline-go/issues/28
  > Is this project no longer developed / maintained or EOL? (since there has been no actions since quiet some time now)
  > yes that's correct, the replacement is here: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azcore

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x104d12484]

goroutine 264 [running]:
github.com/Azure/azure-pipeline-go/pipeline.(*pipeline).newPolicies(0x14000609b40, {0x10581f280, 0x14000018790})
	/Users/js/go/pkg/mod/github.com/!azure/azure-pipeline-go@v0.2.3/pipeline/core.go:155 +0x124
github.com/Azure/azure-pipeline-go/pipeline.(*pipeline).Do(0x14000676080?, {0x10582adf0, 0x14000616b90}, {0x10581f280?, 0x14000018790?}, {0x14000676088?})
	/Users/js/go/pkg/mod/github.com/!azure/azure-pipeline-go@v0.2.3/pipeline/core.go:129 +0x38
github.com/Azure/azure-storage-blob-go/azblob.containerClient.ListBlobFlatSegment({{{{0x14000676080, 0x5}, {0x0, 0x0}, 0x0, {0x14000676088, 0x16}, {0x1400067609e, 0x19}, {0x0, ...}, ...}, ...}}, ...)
	/Users/js/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.14.0/azblob/zz_generated_container.go:596 +0x390
github.com/Azure/azure-storage-blob-go/azblob.ContainerURL.ListBlobsFlatSegment({{{{{...}, {...}, 0x0, {...}, {...}, {...}, 0x0, 0x0, {...}, {...}, ...}, ...}}}, ...)
	/Users/js/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.14.0/azblob/url_container.go:238 +0xe8
go.gazette.dev/core/broker/fragment.(*azureBackend).List(0x1060550a0, {0x10582adf0, 0x14000616b90}, {0x140000408c0, 0x3d}, 0x14000842240?, {0x1400038c730, 0x4e}, 0x140005f4a80)
	/Users/js/Documents/estuary/gazette/broker/fragment/store_azure.go:150 +0x384
go.gazette.dev/core/broker/fragment.List({0x10582adf0, 0x14000616b90}, {0x140000408c0, 0x3d}, {0x1400038c730, 0x4e}, 0x0?)
	/Users/js/Documents/estuary/gazette/broker/fragment/stores.go:136 +0x170
go.gazette.dev/core/broker/fragment.WalkAllStores({0x10582adf0, 0x14000616b90}, {0x1400038c730, 0x4e}, {0x1400093a020, 0x1, 0x0?})
	/Users/js/Documents/estuary/gazette/broker/fragment/index.go:212 +0x11c
go.gazette.dev/core/broker.fragmentRefreshDaemon(0x140000160d0, 0x14000b0e780)
	/Users/js/Documents/estuary/gazette/broker/replica.go:86 +0x1fc
created by go.gazette.dev/core/broker.NewService.func1
	/Users/js/Documents/estuary/gazette/broker/service.go:41 +0x9c
```

I don't love the idea of rewriting code that should work, but this appears to be happening inside a library that is no longer maintained, so I'm somewhat doubtful of whether it'll be able to work. Thoughts?

And the new libraries are clearly not API-compatible with the ones we're using here.

broker/fragment/store_azure.go
jgraettinger: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

Go fmt doesn't strip newlines as Rust's does. It's better than nothing, but Go's formatter was the first kid on the block in a new era of code formatters becoming dejour, and I believe for that reason it's less aggressive, opinionated, and configurable than formatters have become.

Other good material for Go:
* https://go.dev/doc/effective_go
* https://github.com/golang/go/wiki/CodeReviewComments

The `var` thing is a Gazette and former Arbor-ism. At one point I'd put together a whole style guide, but I'm afraid that's lost to time and private repos I don't have access too, and at this point it's institutional knowledge.

jshearer:
Okay @jgraettinger I think I'm ready for another round of 👀 on this. Sorry for all the back-and-forth here, hopefully next time I PR into Gazette it'll be a good bit smoother 😅 

It turns out I was wrong and we absolutely do need `getAzureStorageCredential()` in order to create an `azblob` Pipeline. I put it back and added some comments that I think should match up with the style you're asking for. 

I also ran into an issue when testing this with the scenario that we'll be using it in, in which the service principal is owned by one tenant but the storage account is owned by another. It looks like even though the Azure SDK is supposed to support cross-tenant API calls [without friction](https://github.com/Azure/azure-sdk-for-go/blob/7d4a3cbaadd5bf9f16322e1e2c673a633f146fb1/sdk/azidentity/client_secret_credential.go#L26C8-L26C8), that doesn't appear to work. 

As a result, it appears to be necessary to know ahead-of-time the ID of the tenant that owns the resource we're trying to access, so I had to add another field to `AzureStoreConfig`: `accountTenantID`, as well as plumbing it through from the storage mappings table which you can see over here: https://github.com/estuary/flow/pull/1099

jshearer:
Also, I was having an issue [here](https://github.com/gazette/core/pull/341/files#diff-4e0f308d9322904af173575dc67e0aa258b8934f11306bde95dbe8a6150f33b7R162-R170) where `errors.Cause(err)` was returning `nil` when `err` was clearly non-nil. This was causing `storeRequestTotal.WithLabelValues` to segfault (and thus obscure the actual error that was happening) because the error value was nil, so I changed it to check explicitly for `errors.Cause` unexpectedly returning nil. Is there something else I should have done here instead?

jshearer:
Ugh. This seemed to be working in that it's writing data to the right account in azure storage, but somehow when trying to read that data back out I'm getting an error about the token's issuer being invalid.

```
<AuthenticationErrorDetail>Issuer validation failed. Issuer did not match.</AuthenticationErrorDetail>
```

It seems that the azure sdk is somehow getting tokens from azure AD's v1 endpoint described [here](https://login.microsoftonline.com/7135f765-7f7f-497d-b0c0-25d9557e8226/.well-known/openid-configuration) with `iss` values of `https://sts.windows.net/7135f765-7f7f-497d-b0c0-25d9557e8226/`, whereas it should be using the v2 endpoint [here](https://login.microsoftonline.com/7135f765-7f7f-497d-b0c0-25d9557e8226/v2.0/.well-known/openid-configuration) and getting a token with `iss` of `https://login.microsoftonline.com/7135f765-7f7f-497d-b0c0-25d9557e8226/v2.0`...

I did some digging, and [other people claim](https://stackoverflow.com/questions/59790209/issuer-in-access-token-from-azure-active-directory-is-https-sts-windows-net-wh) that the fix for this is to set a particular value in the app manifest: `"accessTokenAcceptedVersion": 2`. Supposedly this takes some time to propagate, but I set it ~14h ago and still no luck. So... either this just takes _ages_ to propagate, or I'm doing something wrong but I have no idea for the life of me what it is.

Will check again tomorrow :/

jshearer:
Woohoo! Thanks to @psFried for lending me another set of eyes on this, we figured it out! The problem was that I was generating a token for _our_ tenant (i.e the tenant that owns the service principal), rather than _their_ tenant (the tenant that owns the storage account). _Man_ that was obscure... this is working now!

broker/fragment/store_azure.go
```diff
@@ -183,30 +240,96 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
+			return cfg, nil, err
+		}
+
+		serviceClient, err := service.NewClient(cfg.serviceUrl(), identityCreds, &service.ClientOptions{})
+		if err != nil {
+			return cfg, nil, err
+		}
+		a.svcClient = *serviceClient
+
+		credentials, err = getAzureStorageCredential(identityCreds, cfg.accountTenantID)
```
jgraettinger, line 287: The credential that we're building here -- it seems to be specific to the current `ep` ? It embeds their `accountTenantID` ?

Note that there's just _one_ `azureBackend` instance which is registered globally to handle all fragment storage requests across all `azure://` or `azure-ad://` endpoints. As such it can have client instances and/or tokens which are coupled to our own/owned azure account, but cannot have a client that's specific to one end-user account.

Ideally any end-user-specific clients could be cheaply built from first principles by parsing an endpoint / composing with our global client for our service principal. If the underlying client needs to issue a network request each time, that's probably not reasonable however. Compare with `store_s3` and `s3Client` which keeps a multi-map of clients for each configuration of endpoint and profile.

broker/fragment/stores.go
```diff
@@ -159,8 +159,15 @@ func parseStoreArgs(ep *url.URL, args interface{}) error {
 }
 
 func instrumentStoreOp(provider, op string, err error) {
+	// It appears that `errors.Cause` can sometimes return nil
```
jgraettinger, line 162: nit: switch this to use `errors.Unwrap` instead which is now part of the Go standard library since this was written.

jgraettinger (commented):
One significant question around mixing of end-user-configuration in the global `azureBackend` client. The rest is minor stuff

broker/fragment/store_azure.go
```diff
@@ -183,30 +240,96 @@ func (a *azureBackend) azureClient(ep *url.URL) (cfg AzureStoreConfig, client pi
...
+		a.svcClient = *serviceClient
+		// Create an azblob credential that we can pass to `NewPipeline`
+		credentials, err = azblob.NewSharedKeyCredential(accountName, accountKey)
+		if err != nil {
+			return cfg, nil, err
+		}
+	} else if ep.Scheme == "azure-ad" {
+		// Link to the Azure docs describing what fields are required for active directory auth
+		// https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication-service-principal?tabs=azure-cli#-option-1-authenticate-with-a-secret
+		clientId := os.Getenv("AZURE_CLIENT_ID")
```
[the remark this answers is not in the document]
jshearer: Ug sorry, I swear I looked this over for instances of this. 

It's still not clear though. Is this kosher?

`identityCreds, err := azidentity.NewClientSecretCredential(...)`

broker/fragment/stores.go
jshearer: Oh neat, this is what I was missing. Thanks!

broker/fragment/store_azure.go
jshearer: Yuuup, totally right. The only thing that varies here is the `accountTenantID`, which is the tenant of the customer's Azure AD account. Credentials for a particular tenant tenant will let you access any storage account and container that the app has been granted access to. So I changed this to keep track of a `map[string]pipeline.Pipeline` of pipelines per tenant ID, and `(*azureBackend).azureClient(url)` will use that map to cache its clients.

broker/fragment/store_azure.go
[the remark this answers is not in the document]
jgraettinger: Yea, if one of the variables already exists.

`var identityCreds, err = ... ` is an unambiguous assertion to the reader that neither of these variables exist yet in the current scope. Then, `identityCreds, err := ... ` indicates that one of them may exist already in the scope.

jshearer:
Just to confirm: I tested this with two separate catalogs, both using `azure-ad://`, pointed at two separate storage accounts in two separate tenants, and it worked great both writing and reading

jgraettinger (approved):
LGTM

PR #343, opened 2023-07-26
shard unassign: handle non-PRIMARY assignments which are FAILED

jgraettinger:
The API should remove assignments which are non-PRIMARY but are also FAILED.

Also update `gazctl` to process selected shards in batches, so that each RPC is of bounded size and under the Etcd transaction size limit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/343)


jgraettinger:
Manually tested `gazctl` chunking behavior by locally using smaller chunk sizes.

PR #344, opened 2023-08-11
Fix azure pre-signed urls

psFried:
Azure URL signing was broken. The journal prefix was being duplicated in the URL, which resulted in unusable URLs. Clients would fail with a 403 response when attempting to fetch the fragment.  This removes the extra prefix to hopefully allow signing to work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/344)


psFried:
I've tested this locally and confirmed that I was able to read a fragment using the pre-signed URL.

jgraettinger (approved):
LGTM!

PR #345, opened 2023-08-11
message: Sequencer incrementally tracks pending partial sequences

jgraettinger:
If there are a large number of tracked message producers, Sequencer.HasPending() can become a hot loop since it's very frequently run. It's slow because it walks every journal producer to see if it has a larger offset than the last checkpoint offset.

Instead, maintain a set of *partialSeq which begin a new transactional sequence, and clear each *partialSeq upon dequeue of a sequence or rollback. Or, clear _all_ pending sequences upon a Checkpoint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/345)

### DIFF
--- a/message/sequencer.go
+++ b/message/sequencer.go
@@ -29,9 +29,10 @@ import (
 // sufficient to directly read committed messages. When recovering from a
 // checkpoint, or if a very long sequence or old producer is acknowledged, it
 // may be necessary to start a replay of already-read messages. In this case:
-//  * QueueUncommitted will return QueueAckCommitReplay.
-//  * The client calls ReplayRange to determine the exact offset range required.
-//  * The client must then supply an appropriate Iterator to StartReplay.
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM
I have to admit that I find the idea of using a pointer as a map key deeply confusing. I don't really have a better idea, though, and my testing has shown that it works, even after re-assigning the value behind the pointer. I'm assuming that you didn't want to use the `JournalProducer` as the key because that would involve copying the data? Either way, I think this looks good to merge, and we can talk later if you think there's room for improvement. This might just be one of those Go things

jgraettinger:
Thanks. It works because the allocated `partialSeq` is stable for any given journal producer: we re-use the same allocated structure for a given `JournalProducer`.

Given that, the `map[*partialSeq]struct{}` is essentially just indexing / comparing the integer value of the pointer.

PR #348, opened 2023-08-24
gazctl: re-use shard and journal clients

psFried:
This fixes an old bug in `shards prune` where it failed to use the `--broker.address` argument when creating the client.

In addition, this cleans up most of the places where gazctl subcommands were creating multiple new clients for a single operation, and consolidates down to a single client of each type per program invocation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/348)


jgraettinger (approved):
LGTM % more testing

psFried:
I tested all the subcommands, and everything seems to be working as intended. And `shards prune` got a noticeable speedup due from re-using the same clients.

PR #349, opened 2023-09-06
Azure AD: fix my dumb mistake when refreshing user delegation credentials

jshearer:
We must refresh whenever the credential expires _before_ 1 minute from now, not after.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/349)

### DIFF
--- a/broker/fragment/store_azure.go
+++ b/broker/fragment/store_azure.go
@@ -325,14 +325,18 @@ func (a *azureBackend) buildBlobURL(cfg AzureStoreConfig, client pipeline.Pipeli
 
 // Cache UserDelegationCredentials and refresh them when needed
 func (a *azureBackend) getUserDelegationCredential() (*service.UserDelegationCredential, error) {
-	// We want to make sure we create a new credential well before the existing one expires
-	// So this gives us a 60 second buffer before the credential expires to make a new one.
+	// https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-user-delegation-sas-create-cli#use-azure-ad-credentials-to-secure-a-sas
+	// According to the above docs, signed URLs generated with a UDC are invalid after
+	// that UDC expires. In addition, a UDC can live up to 7 days. So let's ensure that
+	// we always sign URLs with a UDC that has at least 5 days of useful life left in it.
 
-	// ----| NOW |------|NOW+1Min|-----| udcExp |---- No need to refresh
-	// ----| NOW  |-----| udcExp |-----|NOW+1Min|---- Need to refresh
+	// ----| NOW |------|NOW+5Day|-----| udcExp |---- No need to refresh
+	// ----| NOW  |-----| udcExp |-----|NOW+5Day|---- Need to refresh
 	// ----|udcExp|-----|  NOW   | ------------------ Need to refresh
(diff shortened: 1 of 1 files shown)


jshearer:
Now that I think about it, I don't think the 1 minute buffer is actually useful here in the first place: AFAICT, the SAS url generated by `SignWithUserDelegation()` can exceed the UDC (user delegation credential)'s expiration, so long as it was created before the UDC expires. Since we always call `getUserDelegationCredential` to get the UDC... I think we can simplify the check to `a.udcExp.Before(time.Now())`. The only concern is if we're so close to the expiration that it expires in between the check and the call to `SignWithUserDelegation()`... and if that happens, then we generate _one_ bad URL, re-try, refresh, and get a good url. @jgraettinger thoughts?

broker/fragment/store_azure.go
```diff
@@ -327,7 +327,11 @@ func (a *azureBackend) buildBlobURL(cfg AzureStoreConfig, client pipeline.Pipeli
 func (a *azureBackend) getUserDelegationCredential() (*service.UserDelegationCredential, error) {
 	// We want to make sure we create a new credential well before the existing one expires
 	// So this gives us a 60 second buffer before the credential expires to make a new one.
-	if a.udc == nil || (a.udcExp != nil && a.udcExp.After(time.Now().Add(time.Minute))) {
+
```
jgraettinger, line 330: nit: extra newline.

nit: I think at least an hour, perhaps even more would be appropriate. Otherwise you can get a signed fragment with a larger requested TTL that's invalidated far before that actual TTL due to the signing credential itself expiring.

If we wanted to get _really_ fancy then we ought to be checking the expiry in comparison to a stamped TTL, but :shrug:

jgraettinger (commented):
LGTM % comment

jshearer:
I take back what I said in my comment. According to the [docs I found](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-user-delegation-sas-create-cli#use-azure-ad-credentials-to-secure-a-sas):

> The SAS is invalid after the user delegation key expires

broker/fragment/store_azure.go
jshearer: Ok, I bumped it all the way up to 5 days minimum UDC window. Microsoft says it maxes out at 7, do what about generating UDCs that last for 6 days, and refreshing once the cached credential has less than 5d left on its timer, so we're always signing URLs with a credential that has at least 5 days left on the clock?

jgraettinger (approved):
One more sanity check: My presumption is that signing key may have a expiry 5 days in the future, but if i request a signed fragment with a 30 second TTL, then my individual stamped fragment URL will still expire in 30 seconds. Correct?

If so, LGTM

jshearer:
That's correct: Signed URL Exp <= UDC Exp

broker/fragment/store_azure.go
```diff
@@ -325,10 +325,18 @@ func (a *azureBackend) buildBlobURL(cfg AzureStoreConfig, client pipeline.Pipeli
```
jgraettinger, line 325: While we're in here, I've noticed that the stamped URLs have an extra '/' which I believe is due to this formatting in url.Parse ☝️ , right? Can we remove the extra "/"?

PR #350, opened 2023-09-28
gazctl: shard pruning should consider all hints

jgraettinger:
If a shard is in a persistent state where it immediately becomes FAILED after recovery, it's very possible for the oldest recovered backup hints to actually be _newer_ than primary hints.

Then, if we prune without considering the primary hints, the shard is unable to recover from its log.

Fortunately this is easily remedied -- delete the primary hints while leaving backup recovery hints in place -- but better to not do it in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/350)

### DIFF
--- a/cmd/gazctl/gazctlcmd/shards_prune.go
+++ b/cmd/gazctl/gazctlcmd/shards_prune.go
@@ -2,7 +2,6 @@ package gazctlcmd
 
 import (
 	"context"
-	"fmt"
 
 	log "github.com/sirupsen/logrus"
 	"go.gazette.dev/core/broker/client"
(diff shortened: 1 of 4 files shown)


cmd/gazctl/gazctlcmd/shards_prune.go
```diff
@@ -53,29 +52,40 @@ func (cmd *cmdShardsPrune) Execute([]string) error {
...
+			// We require that we see _all_ hints for a shards before we may make _any_ deletions.
+			// This is because shards could technically include segments from any log,
+			// and without comprehensive hints which are proof-positive that _no_ shard
+			// references a given journal fragment, we cannot be sure it's safe to remove.
+			// For this reason, we must track the journals to be skipped, so we can be sure
+			// we don't prune journals that are used by a shard that hasn't persisted hints.
+			if hints != nil && len(hints.LiveNodes) > 0 {
+				foldHintsIntoSegments(*hints, logSegmentSets)
+			} else {
+				skipRecoveryLogs[recoveryLog] = true
```
psFried, line 75: Not something that's changed here, but this just made me curious. If there's non-nil hints, but they don't have any `LiveNodes`, could that indicate that all the files have been deleted, and thus we _should_ prune that journal still? Not proposing that we handle that corner case, but more just checking my understanding.

psFried (approved):
LGTM

cmd/gazctl/gazctlcmd/shards_prune.go
jgraettinger: Hmm, that sounds conceptually right but I don't recall if hinted log replay can function if there are literally no LiveNodes. Possibly it handles it by jumping to the head of the log, or possibly it errors out because it wouldn't have an `(offset, sequenceNo, checksum)` tuple to resume reading from.

But if all data is also deleted, that can also be problematic because there must be written & persisted data that's old enough for the reader's offset to be eligible to jump to the log head.

(To my knowledge there's no actual scenario where `LiveNodes` is intentionally empty.)

PR #351, opened 2023-10-06
allocator: make assignment much much faster

jgraettinger:
When used with a large number of journals, certain graph constructions
can be quite slow, particularly during member scale down.

First update `benchmark_test.go` to separately model scale-up and
scale-down phases of a simulated rolling deployment, which replicates
the conditions we've observed.

Next implement the label "gap" heuristic for the push/relabel algorithm
by tracking the number of nodes at each height < len(nodes), and looking
for zero-ings of counts as an indication that a gap has been created.
Respond to gaps by instantly re-labeling nodes to a larger height that
reconnects them to the network.

On my machine, with current (toy) parameters the original algorithm had
a max/flow round which required >4.5s to completed. Now, the same
network completes in 27ms. This scales up combinatorially with graph
complexity.

Update some tests to use "testing" instead of go-check. Excepting where
tests have been extended, all scenarios tests are essentially unchanged.

Finally, bound the number of items that participate in a single max-flow
network, and compose a final desired assignment solution by solving multiple
independent max flow problems, with disjoint subsets of items and all
members.

This makes the big-O runtime of assignment linear on the number of
items, which generally dominates over members.

Testing:
 * Refinements of scenario tests
 * A lot of local testing using the benchmark included in the package. I locally tuned and scaled up the parameters of the benchmark to ramp up the size of the problem, scaling up to a simulated rolling deployment involving 600 members, 1MM items and 3MM assignments, which completed successfully. At this scale, the bottleneck is Etcd itself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/351)


allocator/sparse_flow_network.go
```diff
@@ -324,19 +345,18 @@ func (fs *sparseFlowNetwork) buildCurrentItemArcs(item int, bound int) []pr.Arc
...
-//   member slots.
-// Intuitively, the Member node will resist having more than its fair share of
-// assignments until sufficient pressure builds within the network to indicate
-// that not all assignments can otherwise be made. At this point the member will
-// allow assignments up to its full capacity.
+// buildMemberArc from member `member` to the sink.
 func (fs *sparseFlowNetwork) buildMemberArc(mf *pr.MaxFlow, id pr.NodeID, member int) []pr.Arc {
 	var c = memberAt(fs.Members, member).ItemLimit()
+	// Constrain to the scaled ItemLimit for our portion of the global assignment problem.
+	c = scaleAndRound(c, len(fs.myItems), len(fs.Items))
```
psFried, line 352: Another check for understanding: I think this line addresses my previous question about respecting item limits when treating this as several smaller networks, yeah?

psFried (approved):
This was interesting, and required a few trips to the push-relabel wikipedia page before I felt like I understood what was going on here. I'm sure that I still don't fully understand all of this, and I left a few questions just to make sure I'm not completely lost.

The gap heuristic makes sense, though, and so does the splitting of the problem into smaller groups of items. And FWIW I didn't notice anything that seemed wrong.

allocator/sparse_flow_network.go
jgraettinger: That's right. Each sub-problem gets a corresponding fraction of the total member capacity.

allocator/allocator.go
```diff
@@ -225,19 +206,44 @@ func removeDeadAssignments(txn checkpointTxn, ks *keyspace.KeySpace, asn keyspac
...
+func solveDesiredAssignments(s *State, desired []Assignment) []Assignment {
+	// Number of items to lump into each invocation of push/relabel.
+	// This is an arbitrary number which is empirically fast to solve,
+	// but is large enough that we're unlikely to see further improvements
+	// from finding a global maximum flow.
+	// Splitting the problem in this way makes it big-O linear on items,
+	// instead of combinatorial over members and items.
+	const itemsPerNetwork = 10000
+
+	// NOTE(johnny): this could trivially be parallelized if needed.
```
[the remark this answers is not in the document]
jgraettinger: Answered below.

Then re: parallelization, `State` is unchanged when building / evaluating the max flow network, so multiple goroutines can work in parallel. I didn't do it yet because it's extra SLOC we don't appear to need, and I didn't know for-sure that the memory impact would be negligible.

PR #354, opened 2023-12-07
fix: Azure fragment store with multiple tenants

jshearer:
I knew that the service credentials would be scoped to tenant IDs, but I did not realize that the `service.Client`s and `UserDelegationCredential`s would also be tenant-scoped. This was causing problems where the credentials of whatever tenant was first seen would be used to sign _all_ `SignGet` URLs, resulting in signature mismatch errors when the blob was owned by a different tenant.

This updates the cache logic to keep track of both the service clients and the user delegation credentials by tenant ID, which should solve the problem.

### DIFF
--- a/broker/fragment/store_azure.go
+++ b/broker/fragment/store_azure.go
@@ -2,6 +2,7 @@ package fragment
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -42,12 +43,19 @@ func (cfg *AzureStoreConfig) containerURL() string {
 	return fmt.Sprintf("%s/%s", cfg.serviceUrl(), cfg.containerName)
 }
 
+type UdcAndExp struct {
+	udc *service.UserDelegationCredential
+	exp *time.Time
+}
+
 type azureBackend struct {
-	clients   map[string]pipeline.Pipeline
-	svcClient service.Client
-	clientMu  sync.Mutex
-	udc       *service.UserDelegationCredential
-	udcExp    *time.Time
+	// This is a cache of configured Pipelines for each tenant. These do not expire
+	clients map[string]pipeline.Pipeline
+	// This is a cache of Azure storage clients for each tenant. These do not expire
+	svcClients map[string]service.Client
+	clientMu   sync.Mutex
+	// This is a cache of URL-signing credentials for each tenant. These DO expire
+	udcs map[string]UdcAndExp
 }
 
 func (a *azureBackend) Provider() string {
@@ -56,14 +64,14 @@ func (a *azureBackend) Provider() string {
 
 // See here for an example of how to use the Azure client libraries to create signatures:
 // https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/service/examples_test.go#L285
-func (a *azureBackend) SignGet(ep *url.URL, fragment pb.Fragment, d time.Duration) (string, error) {
-	cfg, _, err := a.azureClient(ep)
+func (a *azureBackend) SignGet(storageMapping *url.URL, fragment pb.Fragment, d time.Duration) (string, error) {
+	cfg, _, err := a.getAzurePipeline(storageMapping)
 	if err != nil {
 		return "", err
 	}
 	blobName := cfg.rewritePath(cfg.prefix, fragment.ContentPath())
 
-	udc, err := a.getUserDelegationCredential()
+	udc, err := a.getUserDelegationCredential(storageMapping)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


broker/fragment/store_azure.go
```diff
@@ -222,23 +237,100 @@ func getAzureStorageCredential(coreCredential azcore.TokenCredential, tenant str
...
-		cfg.accountTenantID, cfg.storageAccountName, cfg.containerName, cfg.prefix = ep.Host, splitPath[0], splitPath[1], strings.Join(splitPath[2:], "/")
+		cfg.containerName, cfg.prefix = storageMapping.Host, storageMapping.Path[1:]
+	} else if storageMapping.Scheme == "azure-ad" {
+		cfg.accountTenantID, cfg.storageAccountName, cfg.containerName, cfg.prefix = storageMapping.Host, splitPath[0], splitPath[1], strings.Join(splitPath[2:], "/")
+	}
+
+	return cfg, nil
+}
+
+func (a *azureBackend) getAzureServiceClient(storageMapping *url.URL) (client *service.Client, err error) {
```
jgraettinger, line 263: this needs to be guarded by a mutex. you're manipulating `a` here.

broker/fragment/store_azure.go
```diff
@@ -324,7 +403,16 @@ func (a *azureBackend) buildBlobURL(cfg AzureStoreConfig, client pipeline.Pipeli
...
-func (a *azureBackend) getUserDelegationCredential() (*service.UserDelegationCredential, error) {
+func (a *azureBackend) getUserDelegationCredential(storageMapping *url.URL) (*service.UserDelegationCredential, error) {
+	if a.udcs == nil {
+		a.udcs = make(map[string]UdcAndExp)
+	}
+	var cfg, err = parseAzureStorageMapping(storageMapping)
+	if err != nil {
+		return nil, err
+	}
+	var udc, hasCachedUdc = a.udcs[cfg.accountTenantID]
```
jgraettinger, line 414: This, and the initialization above, need to be guarded by a mutex.

Please carefully review everywhere that you're mutating and accessing these maps. It always always needs to be guarded.

I would recommend wrapping it in a brief critical section here which handles initialization and pulls out `udc` & `hasCachedUdc`. Then, below where you upsert, lock there as well.

I would rename `clientMu` to just `mu`, and essentially any time you're touching state attached to `a`, it needs to be within the scope of a briefly-held lock. It's perfectly fine, i'd think, if multiple goroutines each independently fetch credentials and overwrite each other :shrug: so you don't need to hold locks for long.

jgraettinger (changes requested):

broker/fragment/store_azure.go
```diff
@@ -56,14 +64,14 @@ func (a *azureBackend) Provider() string {
 
 // See here for an example of how to use the Azure client libraries to create signatures:
 // https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/service/examples_test.go#L285
-func (a *azureBackend) SignGet(ep *url.URL, fragment pb.Fragment, d time.Duration) (string, error) {
-	cfg, _, err := a.azureClient(ep)
+func (a *azureBackend) SignGet(storageMapping *url.URL, fragment pb.Fragment, d time.Duration) (string, error) {
```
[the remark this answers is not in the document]
jshearer: kk, I'll go with `endpoint`. Even though I know it's a convention, the short variable names make it a lot harder to understand what anything is IMO

broker/fragment/store_azure.go
jshearer: Okay, we talked on a VC and my understanding here is: any area that reads _or mutates_ state, and can be entered from multiple threads/goroutines needs a critical section. I believe I got all of them here.

broker/fragment/store_azure.go
```diff
@@ -222,47 +237,133 @@ func getAzureStorageCredential(coreCredential azcore.TokenCredential, tenant str
...
-		return
+func (a *azureBackend) getAzureServiceClient(endpoint *url.URL) (client *service.Client, err error) {
+	var cfg azureStoreConfig
+
+	if cfg, err = parseAzureEndpoint(endpoint); err != nil {
+		return nil, err
 	}
 
-	var credentials azblob.Credential
+	a.mu.Lock()
```
jgraettinger, line 270: nit: can you arrange so that any constructed instance of `azureBackend` already has all maps initialized? Then you don't need to worry about their being nil.

For example:
```
&azureBackend {
   foo: make(map[string]*service.Client),
   bar: make(map[string]string),
}
```

then you can omit nil checks everywhere.

broker/fragment/store_azure.go
```diff
@@ -222,47 +237,133 @@ func getAzureStorageCredential(coreCredential azcore.TokenCredential, tenant str
...
 		}
 		serviceClient, err := service.NewClientWithSharedKeyCredential(cfg.serviceUrl(), sharedKeyCred, &service.ClientOptions{})
 		if err != nil {
-			return cfg, nil, err
+			return nil, err
 		}
-		a.svcClient = *serviceClient
+
+		a.mu.Lock()
+		a.svcClients[accountName] = *serviceClient
```
jgraettinger, line 301: You're dereferencing (making a shallow copy) here and I don't understand why. Seems like `svcClients` should be holding pointers, since that's what's returned by service.NewClientWithSharedKeyCredential().

broker/fragment/store_azure.go
```diff
@@ -34,20 +35,27 @@ type AzureStoreConfig struct {
...
+}
+
 type azureBackend struct {
-	clients   map[string]pipeline.Pipeline
-	svcClient service.Client
-	clientMu  sync.Mutex
-	udc       *service.UserDelegationCredential
-	udcExp    *time.Time
+	// This is a cache of configured Pipelines for each tenant. These do not expire
+	clients map[string]pipeline.Pipeline
```
jgraettinger, line 53: nit: can these be called `pipelines` and `services` perhaps, here and elsewhere? Or `pipelines` and `clients` ?

It's hard to follow which is which.

broker/fragment/store_azure.go
jshearer: Fixed, good point about `svcClients` (now just `clients`) holding a map of pointers.

broker/fragment/store_azure.go
jshearer: Done! It was only constructed in one place, so easy to update.

PR #356, opened 2023-12-11
recoverylog: relax FSM to allow updating the content of a property

jgraettinger:
In certain situations, if a RocksDB doesn't fully initialize then it may re-create its IDENTITY file upon the next next open attempt.

Currently this is prohibited by FSM but should be relaxed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/356)


psFried (approved):
I guess we found our reason to allow it. LGTM

PR #357, opened 2023-12-12
broker/client: discard async appends of deleted journals

jgraettinger:
There are unavoidable scenarios where a client creates an intention to asynchronously write to a journal (as an AsyncAppend, or an checkpointed ACK intent), but the journal is deleted before the write can land.

Before, AppendService will loop forever awaiting the journal to come back into existence.

After, AppendService will discard writes to deleted or non-existant journals, logging a warning before doing so. This parallels the broker behavior of dropping spooled fragment data upon journal deletion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/357)

### DIFF
--- a/broker/client/append_service.go
+++ b/broker/client/append_service.go
@@ -21,6 +21,12 @@ import (
 // writes block until successfully committed, as opposed to handling errors
 // and retries themselves.
 //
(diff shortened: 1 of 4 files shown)


psFried (approved):
LGTM

PR #358, opened 2023-12-13
take two: relax ErrJournalNotFound only when writing recovered ACKs

jgraettinger:
* Revert prior commit.
* Make ErrJournalNotFound a terminal error for AppendService, because it's essentially never recoverable and should be retried at a higher level.
* Have client.Appender handle insufficient journal brokers, as this _is_ a retry-able case that happens.
* Update consumer to relax the writes of recovered ACK intents. Log a missing journal but allow the session to proceed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/358)


psFried (approved):
LGTM

PR #359, opened 2023-12-14
AppendService should retry on JOURNAL_NOT_FOUND

jgraettinger:
Partially roll back a prior commit which began erroring on JOURNAL_NOT_FOUND. There are allowable data races where a created journal is assigned to a primary broker, whose Etcd watch is currently still unaware that the journal even exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/359)


psFried (approved):
LGTM

PR #360, opened 2024-01-23
SegmentSet.Add allow missing LastOffset

psFried:
Allows adding a "covering" segment to a `SegmentSet`. Given an non-empty set, this permits adding a segment that completely covers a subset of existing segments and has `LastOffset: 0`.  This is needed in order to reduce backup and primary hints into the same set, which is needed by `shards prune` in order to prune recovery logs conservatively.

I realized that the other fix is truly not necessary, because the existing behavior of zeroing the final segment of each set of hints is retained, and serves the same purpose.

With this change, I'm able to successfully dry-run pruning all the shards in Estuary's production environment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/360)

### DIFF
--- a/cmd/gazctl/gazctlcmd/shards_prune.go
+++ b/cmd/gazctl/gazctlcmd/shards_prune.go
@@ -152,11 +152,8 @@ func foldHintsIntoSegments(hints recoverylog.FSMHints, sets map[pb.Journal]recov
 	for _, segment := range segments {
 		var set = sets[segment.Log]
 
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM, but can we also [remove this ](https://github.com/gazette/core/blob/master/cmd/gazctl/gazctlcmd/shards_prune.go#L155-L169) ?

PR #361, opened 2024-01-29
broker/consumer: add CreateRevision to ListResponse

jgraettinger:
CreateRevision is the revision at which the JournalSpec or ShardSpec was created within Etcd, and pairs with the ModRevision which is already surfaced in the API response.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/361)


psFried (approved):
LGTM

PR #365, opened 2024-03-08
broker/fragment: region support for s3 stores

williamhbaker:
Previously s3 stores could only use the region available via a given profile or the AWS application default credentials. This adds the ability to set an alternate region in the endpoint URL for an s3 fragment store for buckets that are in a different region.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/365)


jgraettinger (approved):
LGTM

PR #366, opened 2024-03-12
consumer: bound the size of an error stored as Etcd shard status

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/366)

### DIFF
--- a/consumer/shard.go
+++ b/consumer/shard.go
@@ -225,9 +225,14 @@ func servePrimary(s *shard) (err error) {
 		if err != nil && s.ctx.Err() == nil {
 			log.WithFields(log.Fields{"err": err, "shard": s.FQN()}).Error("servePrimary failed")
 
+			var statusErr = err.Error()
+			if len(statusErr) > MAX_ETCD_ERR_LEN {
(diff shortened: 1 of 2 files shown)


psFried (approved):
I noticed that under some circumstances we might [append these errors to an existing list](https://github.com/gazette/core/blob/188c28b10a1f623bc5528f3032973ae36cecdb65/consumer/protocol/shard_spec_extensions.go#L292), and I'm wondering if there's any way for that to still overflow the max etcd request size. I'm guessing that it's probably OK since you've limited the errors here to well under the 1.5MB max size, but just wanted to mention it in case you can think of any scenarios where we might accumulate many errors in the list.

If we wanted to be extra cautious, we could apply a limit in the `Reduce` function, but IDK if that's necessary.

jgraettinger:
> If we wanted to be extra cautious, we could apply a limit in the Reduce function, but IDK if that's necessary.

It's functionally not required, because FAILED is a terminal state for replica status. The reason for that reduction behavior is that we don't want to lose errors if we're reducing a to-be-applied FAILED status into a current PRIMARY status.

PR #367, opened 2024-03-12
generalize first-write persistence and relax byte-zero read skipping

jgraettinger:
Gazette has long had behavior of immediately persisting a fragment having Begin == 0. This has the effect of "dirtying" the fragment index, which ensures that other brokers understand the journal is not in a pristine state in the event of a loss of Etcd consistency.

Generalize this handling to immediately persist a Fragment if it's the first known write of the index, even if at an offset greater than zero. This better handles cases where a journal has been idle for a while, and all of its fragments have expired out via bucket policy, and suddenly new writes for the journal arrive.

Also slightly relax read offset skips to immediately skip-forward a read from (only) byte zero, to a first available **persisted** fragment. This lets such reads proceed much more quickly, rather than waiting for the six-hour delay that's used to guard against the possibility of raced reads vs writes to the fragment index during topology changes.

Testing:
* New and updated unit tests
* Manual testing using the soak crash test framework in kustomize/, which uses a local Minio as a stand-in for S3.
   * Confirmed that deleting _all_ persisted fragments causes the open spool to immediately persist.
   * They continue to flush at flush-interval boundaries.
   * In all other cases, the current fragment is not persisted.
 * Soak crash tests are passing so far; I'll let them run overnight.

This change is [Reviewable](https://reviewable.io/reviews/gazette/core/367)


psFried (approved):
LGTM

PR #368, opened 2024-03-13
journal client: force using http1.1 when fetching fragments from clou…

psFried:
…d storage

We've seen some unfortunate issues with the http2 client, where certain types of transport failures can cause a connection to be unusable, and yet the client will still re-use it for new http2 streams. Disabling http2 is intended to be a short term fix, until the root cause can be identified and fixed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/368)

### DIFF
--- a/broker/client/reader.go
+++ b/broker/client/reader.go
@@ -3,6 +3,7 @@ package client
 import (
 	"bufio"
 	"context"
+	"crypto/tls"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM % CI

PR #371, opened 2024-04-17
mainboilerplate: use a maximum-size connection flow control window

jgraettinger:
Rely only on stream-level flow control, and allow a single broker transport to multiplex LOTS of journals without a flow control deadlock, even if some are not being actively read.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/371)


psFried (approved):
LGTM

PR #384, opened 2024-06-06
support optional TLS for etcd, brokers, and consumers

jgraettinger:
The approach closely parallels the Etcd certificate authorization
model and can support mutual (client and server) certificate
verification.

In server contexts, for $foo (of broker, consumer):

--foo.server-cert-file / FOO_SERVER_CERT_FILE is a path to a cerificate
  for the server to present to clients.

  This option toggles whether TLS is used. If absent, all other TLS
  settings are ignored.

--foo.server-key-file / FOO_SERVER_KEY_FILE is the corresponding private key.

--foo.server-ca-file / FOO_SERVER_CA_FILE is a trusted certificate authority
  for verification of client certificates.

  If this option is omitted, client certificates are verified if
  presented but are not required.

  If this option is set, client certificates are required and verified.

--foo.peer-cert-file / FOO_PEER_CERT_FILE is a path to a cerificate
  for the server to present to peers in a peer-to-peer client context.

--foo.peer-key-file / FOO_PEER_KEY_FILE is the corresponding private key.

--foo.peer-ca-file / FOO_PEER_CA_FILE is a trusted certificate authority
  for verification of peer server certificates. If not specified,
  the system CA pool is used instead.

In client contexts, for $foo (of etcd, broker, consumer):

--foo.cert-file / FOO_CERT_FILE is a path to a client cerificate to
  present to the server context.

--foo.key-file / FOO_KEY_FILE is the corresponding private key.

--foo.trusted-ca-file / FOO_TRUSTED_CA_FILE is a trusted certificate authority
  for verification of server certificates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/384)

### DIFF
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -27,7 +27,7 @@ env:
 jobs:
   build:
     name: "Build"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: "Checkout"
         uses: actions/checkout@v2
--- a/broker/e2e_test.go
+++ b/broker/e2e_test.go
@@ -2,6 +2,7 @@ package broker
 
 import (
 	"context"
+	"crypto/tls"
 	"io"
 	"testing"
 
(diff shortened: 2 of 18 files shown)


psFried (approved):
The code here looks good.
One thing I noticed is that there's a lot of different certificate and key files that can be required, and I think it would help to have some basic docs on setting up MTLS that can clearly outline what certificates are needed and which arguments they get used with.

PR #386, opened 2024-06-17
Phil/colons in names

psFried:
Updates protocol validations to permit colons (`:`) in journal names, shard ids, and label values. Also updates shard id validation to permit the same characters as for journal names, meaning that `= % :` are all now allowed in shard ids.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/386)


jgraettinger (approved):
LGTM

PR #388, opened 2024-06-20
protocol: LabelSelectors support prefix matches

jgraettinger:
In uses of Gazette, labels often encode a hierarchical namespace such as "foo/bar/baz", and for such labels it's desireable to be able to create a LabelSelector that includes or excludes an entire sub-hierarchy of potential label values, such as including or excluding any label that begins with "foo/bar/".

Extend Label to have a Prefix field which may only be set in the context of a LabelSelector (and is a Validate() error otherwise).

In a LabelSelector context, Prefix instructs selector matching to match any label value which is prefixed by the given selector label.

Introduce a convention of "my-label:prefix" as a special suffix which indicates that prefix matching is desired, and update LabelSelector parsing to round-trip such ":prefix" suffixes.

Update the implementation of the special meta-label "prefix" to be in terms of a "name:prefix" selector label, and back out the bespoke implementation that has until-now been used for journal name prefix matching.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/388)


psFried (approved):
LGTM :rocket:

PR #395, opened 2024-07-11
small bug fix for mutex on store_azure

ddowker:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/395)


jshearer (approved):
LGTM, thanks! Clearly we're not exercising the `azure://` scheme like we should be 🤦

PR #397, opened 2024-09-06
broker/client: surface JOURNAL_NOT_FOUND from RetryReader

jgraettinger:
JOURNAL_NOT_FOUND is generally not a retry-able condition. It was orignally made so with a "shrug", on the premise that users may create journals after they had started their consumers. In practice, I'm not aware of any utility for it.

Also tweak logging to reduce noise in environments that have many journals with no current append volume, where read re-authorizations are somewhat frequently obtained.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/397)

### DIFF
--- a/broker/client/reader.go
+++ b/broker/client/reader.go
@@ -26,9 +26,9 @@ import (
 // seek to the requested offset, and read its content.
 //
 // Reader returns EOF if:
-//  * The broker closes the RPC, eg because its assignment has change or it's shutting down.
-//  * The requested EndOffset has been read through.
-//  * A Fragment being read by the Reader reaches EOF.
+//   - The broker closes the RPC, eg because its assignment has change or it's shutting down.
+//   - The requested EndOffset has been read through.
+//   - A Fragment being read by the Reader reaches EOF.
 //
 // If Block is true, Read may block indefinitely. Otherwise, ErrOffsetNotYetAvailable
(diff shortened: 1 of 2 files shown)


psFried (approved):
IIRC there's still a reason for considering `JOURNAL_NOT_FOUND` a retry-able error for appends, due to a race condition when creating new partitions. But this is just for reads, and I also can't think of a reason why we should retry them.

jgraettinger:
At least within Estuary, we create the journal and block until the journal is known to exist (it comes through in a List RPC watch) before we attempt an append to it.

However, now you've got me worried about a data-race where the broker serving and responding to the List RPC sees the journal creation _just slightly before_ the broker to which an Append RPC is directed, and the Append doesn't have a read-through revision attached.

Perhaps a more-conservative policy is to error iff we _had_ a Read RPC that was open and server-closed, and then next see a JOURNAL_NOT_FOUND

jgraettinger:
Now I'm worried about a creation race condition where the broker serving a List RPC sees the journal creation just slightly ahead of the broker to which the Read RPC is directed (the read doesn't have a read-through revision attached).

jgraettinger:
PTAL -- we'll now make a few attempts if the request doesn't have a read-through revision, and _then_ error.

PR #406, opened 2024-10-17
message: account for duplicates across a recovered checkpoint boundary

jgraettinger:
Sequencer's runtime assertion checks were slightly too strong. A scenario we encountered which tripped a runtime panic was:

- CONTINUE_TXN messages were read from a journal, and were then part of a checkpoint. The checkpoint included the producer's earlier last ACK clock and a begin offset at the first of these messages.

- The messages were then duplicated within the journal, and ACKed.

- A new Sequencer recovers from the checkpoint.

- It reads the later, duplicated messages and adds them to the ring, followed by an ACK which begins a dequeue with a replay-read of the earlier messages.

- The earlier messages are dequeued first, and then the replay ends and hands off to the ring.

- Sequencer blows up because the first ring message is not strictly larger than the largest Clock dequeued from the replay read.

The fix is simply to remove the runtime assertion, and discard duplicates, as this is valid thing that can happen. Also add a new test case covering this scenario.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/406)

### DIFF
--- a/message/sequencer.go
+++ b/message/sequencer.go
@@ -401,40 +401,21 @@ func (w *Sequencer) Step() error {
 		w.dequeuedClock = GetClock(uuid)
 
 		if w.dequeuedClock != 0 && w.dequeuedClock <= w.emit.minClock {
-			if w.replayIt != nil {
-				continue // These can happen during replays.
-			} else {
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

PR #407, opened 2024-11-19
keyspace: Etcd watch can retry on Unknown gRPC errors

jgraettinger:
We should continue to poll the watch channel after Unknown errors rather then bailing out with a terminal error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/407)


psFried (approved):
LGTM

PR #408, opened 2024-11-26
update etcd dependency

psFried:
This pulls in the fix for an issue that could cause the etcd client to use an incorrect revision when re-starting a watch after an etcd server failure. This presented as a crash in production due to receiving a watch event with a revision equal to (not greater) than the last processed revision. It was difficult to follow the discussion in the etcd issues, but this comment sums it up nicely:
https://github.com/etcd-io/etcd/pull/15248#issuecomment-1420125977

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/408)


jgraettinger (approved):
LGTM

PR #414, opened 2025-01-16
broker: tear down a stale append pipeline before proxying

jgraettinger:
This can happen if we were primary, but have been re-assigned as a replica without an observed cancellation of our assignment in between.

We must tear down the stale pipeline to release replica spools before we return it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/414)


psFried (approved):
Looks good

PR #416, opened 2025-01-17
scale allocator member ItemLimit to the relative share of ItemSlots, and refine suspension eligibility

jgraettinger:
ItemSlots may be unevenly distributed amoung assignment sub-problems.

We need to account for this, by scaling member ItemLimits by the relative portion of assignment slots that must be allocated within each sub-problem, rather than just the relative number of items.

Also, tune journal suspension to apply the journal's configured flush interval when evaluating whether the index is "dirty".

See individual commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/416)


psFried (approved):
LGTM

PR #422, opened 2025-03-02
message: nil out dequeued Envelope.Message after consumption

jgraettinger:
Otherwise, the ring will indefinetely retain messages which have already been processed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/422)


dyaffe (approved):
LGTM

PR #424, opened 2025-03-14
shard: convert errors to runes before truncating

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/424)


consumer/shard.go
```diff
@@ -225,14 +225,13 @@ func servePrimary(s *shard) (err error) {
 		if err != nil && s.ctx.Err() == nil {
 			log.WithFields(log.Fields{"err": err, "shard": s.FQN()}).Error("servePrimary failed")
 
-			var statusErr = err.Error()
-			if len(statusErr) > MAX_ETCD_ERR_LEN {
-				statusErr = statusErr[:MAX_ETCD_ERR_LEN] + " ...[truncated]"
+			var err = []rune(err.Error())
```
psFried, line 228: just to double check: What happens if this error string already contains an invalid utf8 sequence? I'm not 100% certain whether that's even possible, but maybe?

psFried (approved):
LGTM

consumer/shard.go
jgraettinger: Yep, excellent question: https://go.dev/play/p/cNv5eZBTtbA

jgraettinger: Adding a comment to this effect, as I had to research it as well to ensure it wouldn't panic.

PR #429, opened 2025-04-14
consumer: Apply RPC takes optional hints to set on upsert

jgraettinger:
Certain migration use-cases require that shards new to a cluster also have pre-existing FSM hints available for their recovery log.

Piggy-back optional FSM hints onto the Apply RPC which creates the shard, so that they're created in the same Etcd transaction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/429)


consumer/protocol/rpc_extensions_test.go
```diff
@@ -127,6 +130,8 @@ func (s *RPCSuite) TestApplyRequestValidationCases(c *gc.C) {
 	req.Changes[1].Delete = "other-valid-id"
 	c.Check(req.Validate(), gc.ErrorMatches, `Changes\[1\]: invalid ExpectModRevision \(0; expected > 0 or -1\)`)
 	req.Changes[1].ExpectModRevision = 1
+	c.Check(req.Validate(), gc.ErrorMatches, `Changes\[1\]: hints may be set only with an upsert, not a delete`)
```
psFried, line 133: I'd expect to see this same validation error for the change at index 0, since the validation checks whether both `Delete` and `PrimaryHints` are set.

consumer/protocol/rpc_extensions.go
```diff
@@ -102,6 +102,8 @@ func (m *ApplyRequest_Change) Validate() error {
 			return pb.ExtendContext(err, "Delete")
 		} else if m.ExpectModRevision <= 0 && (m.ExpectModRevision != -1) {
 			return pb.NewValidationError("invalid ExpectModRevision (%d; expected > 0 or -1)", m.ExpectModRevision)
+		} else if m.PrimaryHints != nil {
```
psFried, line 105: What about validation of the hints themselves? I'm not actually sure to what extent we really _can_ validate them, as long as they parse OK, especially in this function.

consumer/shard_api.go
```diff
@@ -126,6 +127,14 @@ func ShardApply(ctx context.Context, claims pb.Claims, srv *Service, req *pc.App
 			}
 			key = allocator.ItemKey(s.KS, change.Upsert.Id.String())
 			ops = append(ops, clientv3.OpPut(key, change.Upsert.MarshalString()))
+
+			if change.PrimaryHints != nil {
```
psFried, line 131: optional suggestion: it might be worth logging something here

IDK if there's any other validation we should be doing on the hints, but if we wanted to check the actual journals and offsets, then we'd need to do that here. IDK how much value there is in that, though, since if they're wrong we'll find out pretty quick once the shard starts.

psFried (approved):
LGTM % a few questions

consumer/protocol/rpc_extensions_test.go
jgraettinger: delete and upsert raises a different error, and test code above clears the `delete` field so it's unset at this point.

consumer/shard_api.go
jgraettinger: Yea, beyond running Validate (which I've added), I think that's out of scope for this API.

PR #434, opened 2025-06-12
Update Apply APIs to authorize Upserts against full label sets

jgraettinger:
This change updates both the broker and consumer Apply APIs to handle label selectors differently for Upserts and Deletes:

For Upserts:
- Authorization now checks against the full label set (including meta-labels)
- Uses LabelSetExt() to get all labels including "name" (broker) or "id" (consumer)
- Ensures consistency with how List operations authorize

For Deletes:
- Maintains backward compatibility with name-only (broker) or id-only (consumer) authorization
- No change in behavior for delete operations

This change ensures that users with label-based claims can only upsert journals/shards if their claims match all the labels on the resource, not just the name/id. This provides more granular access control.

🤖 Generated with [Claude Code](https://claude.ai/code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/434)

### DIFF
--- a/broker/list_apply_api.go
+++ b/broker/list_apply_api.go
@@ -219,30 +219,26 @@ func (svc *Service) Apply(ctx context.Context, claims pb.Claims, req *pb.ApplyRe
 		return nil, err
 	}
 
-	// The Apply API is authorized exclusively through the "name" label.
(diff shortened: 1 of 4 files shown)


psFried (approved):
LGTM

PR #435, opened 2025-06-17
fix two rare bugs observed in Etcd response handling

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/435)


psFried (approved):
LGTM

PR #437, opened 2025-06-25
fragment: slightly relax offset skip behavior

jgraettinger:
Currently we will immediately skip forward a read if it begins at offset zero.

However, we also see cases where a long-disabled task will come back online, observe an ACK, and then attempt a replay-read from an offset fixed in its recovery log. These reads block for the full offset jump threshold.

Resolve this by skipping forward any read having an offset behind the minimum available offset of journal content.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gazette/core/437)


psFried (approved):
LGTM

PR #445, opened 2025-08-13
server: offer "http/1.1" and "h2" via ALPN

jgraettinger:
As per the gRPC spec, recent clients require that we offer "h2".

So, offer both "http/1.1" and "h2", preferring "http/1.1". Regular HTTP clients will negotiate http/1.1 (h2 is not supported), and gRPC clients (which don't offer http/1.1) will negotiate h2.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -108,6 +108,15 @@ func New(
 			grpc.MaxRecvMsgSize(int(maxGRPCRecvSize)),
 		),
 	}
+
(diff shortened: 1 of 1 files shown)


jgraettinger:
The standard library server doesn't support HTTP/2 today given how `cmux` is used -- it cannot cast to `*tls.Conn` and thus will not take the HTTP/2 codepath (which depends on observing `h2` in the ALPN resolution).

PR #448, opened 2025-09-15
message: change Clock.Tick() to increment by 1 microsecond

jgraettinger:
Update Clock.Tick() to increment by 1 microsecond (1000ns) instead of 100ns to improve compatibility with external systems that only have microsecond precision. This changes the increment from 1 to 160 units in the clock's internal representation (10 * 2^4 bits).

An implication is that a _single_ producer can write at most one million messages per second (1M micros in a second). Multiple producers can of course write at a higher rate.


williamhbaker (approved):
LGTM

PR #461, opened 2026-02-27
broker: size read chunks to fit gRPC's 32KB buffer pool tier

jgraettinger:
Reduce chunkSize from 128KB to (32KB - 32) so that marshaled ReadResponse messages stay under 32KB. gRPC's internal buffer pool has tiers at 256, 4KB, 16KB, 32KB, and 1MB — exceeding 32KB rounds up to 1MB allocations which can be pinned for extended periods when readers are slow, causing excessive memory usage.

### DIFF
--- a/broker/read_api_test.go
+++ b/broker/read_api_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"math"
 	"net/url"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


broker/read_api_test.go
```diff
@@ -496,6 +497,20 @@ func TestReadRequestErrorCases(t *testing.T) {
...
+		Offset:    math.MaxInt64,
+		WriteHead: math.MaxInt64, // Not actually set with content chunks.
+		Content:   make([]byte, chunkSize),
+	}
+	require.Less(t, resp.ProtoSize(), 1<<15) // 32,768 bytes.
+
+	var buf, _ = resp.Marshal()
+	require.Equal(t, len(buf), resp.ProtoSize())
+	require.Equal(t, len(buf), 32760)
+
```
williamhbaker, line 511: nit: extra line here

williamhbaker (approved):
LGTM

PR #463, opened 2026-03-16
Revert "allocator: zero limit in SetExiting for backward compatibility"

williamhbaker:
This reverts commit a0705223020b8b2dfb92e63c226a473d181c66d3.

Allocator deployments going forward must first have been running release v0.102.0 prior to deploying on this commit.

I did a fresh round of manual tests here to verify that the transition from brokers on `v0.102.0` to this one works as it should: Old brokers can stop, and these new ones can replace them. Also when all brokers run this new version, they can be started and stopped individually, or as a rolling deployment; or to simulate the unattended upgrades case they can all be restarted at the same time and end up taking turns restarting.


jgraettinger (approved):
LGTM!

PR #466, opened 2026-04-07
broker: size List response chunks to fit gRPC's 32KB buffer pool tier

williamhbaker:
Reduce `maxJournalsPerListResponse` from 1000 to 25 so that marshaled `ListResponse` messages stay within the 32KB gRPC buffer pool tier.

At typical journal sizes (~750-1300 bytes per `ListResponse_Journal`), 25 journals produces messages of ~19-32KB. Previously, larger chunks produced messages > 32 KB, landing in the 1MB buffer pool tier. When many List watch streams sent responses concurrently, this caused significant memory spikes from pooled 1MB buffers.


jgraettinger (approved):
LGTM

PR #468, opened 2026-04-30
gazctl: shards prune improvements

williamhbaker:
Skip attempting to prune suspended recovery log journals, since they have no fragments list available. There is still an open question of how these suspended recovery log journals should get pruned, since they may contain prunable data, but that is left as a separate exercise. For now, this will just keep the shards prune command from failing fatally any time there is a suspended recovery log journal for the shards matching the selector.

Also, improve execution time by parallelizing fragment removals. A simple `errgroup` directly in the `shards prune` machinery was used, rather than adding a bulk removal method to the `Store` interface. This was a practical choice, since GCS stores don't have a bulk deletion endpoint anyway.

### DIFF
--- a/broker/client/list.go
+++ b/broker/client/list.go
@@ -252,6 +252,8 @@ func ListAllFragments(ctx context.Context, client pb.RoutedJournalClient, req pb
 			return resp, mapGRPCCtxErr(ctx, err)
 		} else if err = r.Validate(); err != nil {
 			return resp, err
+		} else if r.Status == pb.Status_SUSPENDED {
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

jgraettinger:
> There is still an open question of how these suspended recovery log journals should get pruned, since they may contain prunable data, but that is left as a separate exercise.

By definition, if ErrSuspended is being returned it means that the journal has no fragments. Otherwise, it would only be partially suspended and actual fragments would be returned by the ListFragments RPC.

PR #469, opened 2026-05-01
broker: add MinEtcdRevision to ReadRequest and AppendRequest

williamhbaker:
Lets clients require the broker's `KeySpace` to have caught up to a known Etcd revision before resolving their request. Avoids races where a client that learned of a journal at revision `N` sends a request to a broker whose watch is still behind `N`.

Related: https://github.com/estuary/flow/pull/2912


jgraettinger (approved):
LGTM

PR #470, opened 2026-06-01
broker: rebuild replica when journal is deleted then re-created

jgraettinger:
When a journal name is deleted and re-created such that both Etcd events coalesce into a single KeySpace.Apply batch, the journal never appears to leave LocalItems. The resolver would then keep the prior incarnation's replica, resurrecting its stale spool/index rather than starting the new journal empty.

Track each replica's Etcd CreateRevision and rebuild (tearing down the old) when it changes, since CreateRevision changes only on delete-then- create and not on ordinary spec updates.


williamhbaker (approved):
LGTM

PR #471, opened 2026-06-02
broker: add SUSPEND_KEEP for reset-head of suspended journals

jgraettinger:
reset-head previously issued its zero-byte append with the zero-value SUSPEND_RESUME, which silently resumed a partially-suspended journal -- reverting its Suspend level from PARTIAL to NONE and forcing a costly restoration of the full replication topology -- merely to reset the write head.

Add AppendRequest_SUSPEND_KEEP, which proceeds without altering the journal's suspension level: a partially-suspended journal is reset through its single remaining replica and stays suspended, while a fully-suspended journal still fails with SUSPENDED (it has no replica to serve the append). Like other non-default suspend controls, a SUSPEND_KEEP append may carry only an empty barrier; content is rejected with NOT_ALLOWED, as it would otherwise be committed through the journal's reduced-replication topology.

gazctl journals reset-head now issues SUSPEND_KEEP, so resetting the head of a suspended journal no longer disturbs its suspension.


williamhbaker (approved):
LGTM

PR #472, opened 2026-06-03
keyspace: restart watch when channel closes with a live context

williamhbaker:
The Etcd client usually provides a causal error when a watch is terminated, like `ErrNoLeader`, but has a 250ms timeout associated with reading the error from the channel, after which the channel is simply closed.

On watch channel closure, in the absence of any `ctx.Err()`, we must consider the watch failed and restart it to prevent an allocator leader being unable to observe updates in rare circumstances.


jgraettinger (approved):
LGTM

PR #481, opened 2026-06-03
broker: fix data race from leaked goroutine in fragment store health test

jgraettinger:
TestFragmentStoreHealthCases created a broker via newTestBroker but never called cleanup(), so its resolver watch goroutine outlived the test. When a subsequent test called SetSharedPersister, the leaked goroutine was still reading the sharedPersister global in newReplica, tripping the race detector under `go test -race`.

Add the missing `defer broker.cleanup()` to tear down the broker's tasks.Group at end of test, matching every other broker test.


williamhbaker (approved):
LGTM

PR #488, opened 2026-07-15
gazctl shards prune: treat missing delete permission as non-fatal

williamhbaker:
Recovery-log pruning is best-effort cleanup of superseded fragments. When a tenant brings their own storage bucket without granting delete permission, every prune fails with AccessDenied on the recovery/ prefix. This hard-errors the command (exit 1) and fires recurring alerts for a bucket we can't clean up and don't otherwise depend on.

Classify a failed fragment removal: authorization failures (missing delete permission, 403, bucket-not-found) are now logged and counted in a new non-fatal permissionDenied metric, so the prune still completes successfully. All other removal errors retain the existing failedToRemove behavior and continue to fail the command, so genuine problems keep surfacing.

This reuses the per-backend IsAuthError classification already used by fragment.Persist for graceful shutdown, exposed via a new fragment.IsAuthError helper.

### DIFF
--- a/broker/fragment/stores.go
+++ b/broker/fragment/stores.go
@@ -111,6 +111,16 @@ func Remove(ctx context.Context, fragment pb.Fragment) error {
 	return store.Remove(ctx, fragment.ContentPath())
 }
 
+// IsAuthError reports whether `err`, as returned by an operation such as Remove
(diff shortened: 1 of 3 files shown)


dgreer-dev (approved):
Looks good.

PR #489, opened 2026-07-15
broker: add opt-in delete-permission probe to FragmentStoreHealth

williamhbaker:
The storage-mapping connection test never exercises a delete, so a bucket policy missing the delete action passes setup and only fails later, when recovery-log pruning hits AccessDenied and fires recurring alerts.

Add check_delete and check_delete_prefix to FragmentStoreHealthRequest. When check_delete is set, the broker writes and removes a throwaway probe object under the prefix (empty probes the store root) to verify delete permission at configuration time. The probe is on-demand and excluded from the periodic background check, so existing stores lacking delete permission keep reporting healthy. It uses the generic Store interface, so S3, GCS, and Azure are all covered.

Note: Manually tested locally against an S3 emulator.


dgreer-dev (commented):
Since check_delete makes the broker do a Put/Remove against the store, should this require more than Capability_READ?

williamhbaker:
I was thinking about that but we are already doing a PUT (write) as part of the normal health checks so it didn't seem like  much more of a deviation than that.

dgreer-dev (approved):
LGTM

PR #493, opened 2026-08-24
gazctl shards prune: retry transient fragment removal errors

williamhbaker:
Object deletes fail intermittently with errors which resolve on their own: 5xx responses, rate limits, and dropped connections. The GCS client does not retry a delete having no preconditions, so a single such error counts as a failed removal and fails the whole prune with exit 1, alerting on cleanup which would have succeeded on a second attempt.

Retry a failed removal up to three times, a second apart. Authorization failures are returned as before, without retry, and remain counted as a non-fatal permissionDenied.

There's a known gap with this implementation where on GCS a deletion might actually succeed but the success response gets lost coming back to us, and then we retry and get a 404 and fail. I haven't seen that actually happen yet conclusively, so I'll worry about that later.


dgreer-dev (approved):
LGTM.  Is a second too short for the retry?

williamhbaker:
> Is a second too short for the retry?

I'm assuming these are very transient errors where a retry would immediately resolve. It's possible that 1 second would be too short though if it's something other than that. If we see the failures continue I'll tune it; hopefully it isn't an issue though.

In the somewhat near future I'm planning on making larger changes to our recovery log / pruning architecture in Flow so just trying to get by until then.

PR #495, opened 2026-09-09
store-rocksdb: return barrier errors from the sync hook

williamhbaker:
Sync, Fsync and RangeSync discarded the recorder barrier error, so RocksDB reported a write as synced when its operations never committed to the log. A consumer could then publish ACKs for a checkpoint which cannot be recovered.

These methods now return error. The cgo callbacks return an owned C string which the C++ hook converts to Status::IOError, failing the write. RocksDB latches the failed sync, so later writes also fail.

The first commit is a clean revert of https://github.com/gazette/core/pull/482, which was generally insufficient and is no longer needed with the explicit error return.


jgraettinger (approved):
LGTM

